PR #43, opened 2018-07-31
Adds SubscriptionRead and ISubscriptionBridge

marceloschmidt:


graywolf336:
[subscription-testing_0.0.12.zip](https://github.com/RocketChat/Rocket.Chat.Apps-engine/files/4610615/subscription-testing_0.0.12.zip)

Here is the app to run to test. It simply logs the subscription's of a room to the logger (aka don't do it on a big room otherwise the logs will be filled).

src/definition/subscriptions/IRoomSubscriptionIterator.ts
```diff
@@ -0,0 +1,3 @@
+import { IRoomSubscription } from './IRoomSubscription';
+
+export interface IRoomSubscriptionIterator extends AsyncIterableIterator<IRoomSubscription> {}
```
graywolf336, line 3: I may get rid of this.

tsconfig.json
```diff
@@ -12,7 +12,7 @@
         "experimentalDecorators": true,
         "moduleResolution": "node",
         "types": ["node"],
-        "lib": ["es2017", "dom"],
+        "lib": ["es2018", "dom"],
```
lolimay, line 15: Is it on purpose 🧐?

graywolf336: Yes sir. Rocket.Chat's version of NodeJS got updated to v12 :)

Issue #46, opened 2018-08-23
[BUG] Unable to start RocketChat under windows


twi123:
Unable to run RocketChat under Windows due exception:
```
...
 Exception in callback of async function: Error: Could not find the Apps TypeScript Definition Package Version!
...
```
because of [AppPackageParser.ts@151](https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/7bf4cc158f66c1a99d60ec5b1a7ee704e3c75176/src/server/compiler/AppPackageParser.ts#L151)
```
...
        const prodLocation = __dirname.split('@rocket.chat/apps-engine')[0] + '@rocket.chat/apps-ts-definition/package.json';
...
```

Splitting path by substring contains **/** - bad idea.

thassiov:
Closing issue as user already states that the problem is fixed
https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/47#issuecomment-425118948

PR #47, opened 2018-08-23
[FIX] Issue #46: replace / with path.sep

twi123:
Fixes Issue #46 

Replace `/` with `path.sep` for working with paths under Windows


rodrigok:
@sorry by the delay on dealing with this pull request.

We've changed somethings here and the code you touched is not the same anymore, can you check if the new code still have the same issue? And if yes can you resolve the conflict here to submit your solution again?

Thanks

twi123:
Yes, works like a charm!
Thanks!

> @twi123 can you check my previews comment? I'll close this PR meanwhile.

Yes, works like a charm!
Thanks!

Issue #56, opened 2018-08-29
Allow Dependencies on npm Packages


rodrigok:
@graywolf336 commented on [Tue May 15 2018](https://github.com/RocketChat/Rocket.Chat.Apps-ts-definition/issues/49)

Currently Apps can not bring in third party packages in a supported manner (yes there are ways around it), however this is a big need and if we don't allow this then Apps will be severely limited. However, just allowing Apps to include any and every npm package will open the door to security issues and even malware packages.

So, we need a way to provide "secure" way for Apps to define the npm packages they use. This will be a multi-step process. First up, we will add a `npmDepends` configuration section of the `App.json` configuration file. Then whenever someone submits an App to the Marketplace we will look through that section and ensure they seem to be trusted modules. After it has been marked as reviewed and "trusted", an automated system will package in the npm modules (we will need to look into the license restrictions here) into the App.

However, there will be Apps that want to not distribute themselves via the Marketplace for various reasons (let's say an in-house App). Due to this, we will provide a way for Apps to package the npm modules themselves but a big, not so nice, warning will be displayed when installing and enabling the App that warns of potential risks associated with that App.

---

I would like the thoughts of any and everyone on this manner. Does this make sense and is it logical to do this? Yes, the Apps Marketplace will already have a manual review process so adding one more task to it just makes sense.

---

@timkinnane commented on [Wed May 16 2018](https://github.com/RocketChat/Rocket.Chat.Apps-ts-definition/issues/49#issuecomment-389695758)

Hey Bradley, on a technical note, is `npmDepends` a common approach or specific to Meteor? My understanding is that the apps will not be Meteor packages, so why not just use standard `package.json` manifest? Seems like reinventing the wheel otherwise?

Generally, I would err on the side of less restriction rather than the tight controls you're suggesting. Every node package is just the tip of its own tree of dependencies, there's no feasible way to enforce security standards for all the code in use, so it seems arbitrary and unnecessarily restrictive to enforce oversight at that level. Not to mention an intensive task for manual reviewers.

I'd suggest just listing the included dependencies on the app's marketplace profile, perhaps with some stats for each package, e.g: number of downloads, number of dependents, up to date dependencies (david-dm.org badge). There's services available to link to or integrate, for auditing an NPM package too, e.g. [packagequality](http://packagequality.com/#?package=@rocket.chat%2Fsdk).
[...]

PR #61, opened 2018-09-03
[NEW] Api provider

### LINKED ISSUE #12: Implement the Webhooks
Allow endpoints for Rocket.Chat Apps which will be fully functioning webhooks which support get, post, put, delete, etc.

TypeScript Definition: https://github.com/RocketChat/Rocket.Chat.Apps-ts-definition/tree/master/src/webhooks

rodrigok:
Closes #12

Rocket.Chat bridge implementation https://github.com/RocketChat/Rocket.Chat/pull/11938

TODO
- [x] List provided webhooks on app's details page to be easy to copy
- [x] Add property to define hook as private or public
- [x] Add property to define hook security level
- [x] Property to provide curl samples for each hook

IMPROVEMENTS
- [ ] Implement security level TOKEN_AUTH
- [ ] Implement security level SHARED_SECRET

Webhook admin UI
image

Implementation example
```typescript
// index.ts
import {
    IConfigurationExtend,
    ILogger,
import { ApiSecurity, ApiVisibility } from '@rocket.chat/apps-engine/definition/api/IApi';

import { MessageEndpoint } from './endpoints';

export class JitsiApp extends App {
    constructor(info: IAppInfo, logger: ILogger) {
        super(info, logger);
    }

    protected async extendConfiguration(configuration: IConfigurationExtend): Promise<void> {
        await configuration.api.provideApi({
            visibility: ApiVisibility.PRIVATE,
            security: ApiSecurity.UNSECURE,
            endpoints: [
                new MessageEndpoint(),
            ],
        });
    }
}
```
```typescript
// endpoints.ts
import { HttpStatusCode, IHttp, IModify, IPersistence, IRead } from '@rocket.chat/apps-engine/definition/accessors';
import { IApiRequest, IApiResponse } from '@rocket.chat/apps-engine/definition/api';
import { IApiEndpoint } from '@rocket.chat/apps-engine/definition/api/IApi';
import { example } from '@rocket.chat/apps-engine/definition/api/IApiExample';
import { IApiEndpointInfo } from '@rocket.chat/apps-engine/definition/api/IApiEndpointInfo';

export class MessageEndpoint implements IApiEndpoint {
    public path: string = 'message/:id';

    @example({
        params: {
            id: 'message_id',
        },
        query: {
            asd: '123',
        },
        headers: {
            'X-Auth': 'asd: 123',
        },
    })
    public async get(request: IApiRequest, endpoint: IApiEndpointInfo, read: IRead, modify: IModify, http: IHttp, persis: IPersistence): Promise<IApiResponse> {
        return {
            status: HttpStatusCode.OK,
            headers: {
                'X-Auth': 'asd: 123',
            },
            content: {
                test: request.query.asd,
            },
        };
    }

    @example({
        params: {
            id: 'message_id',
        },
        query: {
            asd: '123',
        },
        content: {
            content: 1,
        },
    })
    public async post(request: IApiRequest, endpoint: IApiEndpointInfo, read: IRead, modify: IModify, http: IHttp, persis: IPersistence): Promise<IApiResponse> {
        return {
            status: HttpStatusCode.OK,
            content: request.content,
        };
    }
}
```


rodrigok:
@graywolf336 can you provide some examples?

piotrkochan:
awesome!

src/server/managers/AppWebhook.ts
```diff
@@ -0,0 +1,108 @@
...
+    'get',
+    'post',
+    'put',
+    'delete',
+    'head',
+    'options',
+    'patch',
+];
+
+export class AppWebhook {
```
graywolf336: What's the purpose of having two `AppWebhook` items within the managers folder?

geekgonecrazy: @rodrigok @graywolf336 renamed from webhook to api.

rodrigok:
@geekgonecrazy awesome, can you remove the docs changes? I'm planning to move docs to a dedicated branch to prevent this mess on PRs as Bradley suggested some time ago.

geekgonecrazy:
@rodrigok done :)

package.json
```diff
@@ -57,19 +56,20 @@
...
     "pre-commit": "^1.2.2",
     "tap-bark": "^1.0.0",
-    "ts-node": "^6.0.2",
-    "tslint": "^5.9.1",
+    "ts-node": "^6.2.0",
+    "tslint": "^5.11.0",
     "typedoc": "^0.11.1"
   },
   "dependencies": {
+    "@types/node": "^10.9.4",
```
graywolf336: Any reason the this has been added to the dependencies?

rodrigok: Yes, was needed to allow the internal module imports, but I found another solution for that. So I'll remove it.

src/definition/api/IApi.ts
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * Provide enpoints for this api registry
+     */
+    endpoints: Array<IApiEndpoint>;
+}
+
+export enum ApiVisibility {
+    /**
+     * Public api's have a fixed URL format, they are easy to remember
+     * but they are easy to be hacked as well.
```
graywolf336: Can we change the `hacked` to `guessed`? Maybe reword this to be:

`A public Api has a fixed format for a url. Using it enables an easy to remember structure, however, it also means the url is intelligently guessed. As a result, we recommend having some sort of security setup if you must have a public api. Whether you use the provided security, ApiSecurity, or implement your own.`

src/definition/api/IApi.ts
```diff
@@ -0,0 +1,82 @@
...
+import { IApiEndpointInfo } from './IEndpoint';
+import { IApiRequest } from './IRequest';
+import { IApiResponse } from './IResponse';
+
+export interface IApiEndpoint {
+    /**
+     * Path to complete the api URL. Example: https://{your-server-address}/api/apps/public/{your-app-id}/{path}
+     * or https://{your-server-address}/api/apps/private/{your-app-id}/{private-hash}/{path}
+     */
+    path: string;
```
graywolf336: What would you think about allowing param style route definitions? For example: `/contacts/:contactId/numbers` would result in the passed in `contactId` being a param. Maybe this can be a future plan for improvements

src/definition/api/IApi.ts
```diff
@@ -0,0 +1,82 @@
+import { IHttp, IModify, IPersistence, IRead } from '../accessors';
+import { IApiExample } from './IApiExample';
+import { IApiEndpointInfo } from './IEndpoint';
+import { IApiRequest } from './IRequest';
+import { IApiResponse } from './IResponse';
+
+export interface IApiEndpoint {
```
graywolf336: I am still of the opinion this should be a class. That way we can have helper methods for converting various things from one to another. For example, the `IApiRequest` has `content: any`. But let's be real, most people are going to expect that to be a json value with a format of `{ [key: string]: string }`. So, we could have helper methods for the TypeScript to assume that something is what they expect it to be. Also, for converting from a string to json. Could also have helper methods for common things such as a successful response `return this.success()` or `return this.json(statusCode: number, data: object)`.

src/server/managers/AppApiManager.ts
```diff
@@ -0,0 +1,156 @@
...
+     *
+     * @param appId the app which is providing the api
+     * @param path the path to be executed in app's api's
+     * @param request the request data to be evaluated byt the app
+     */
+    public executeApi(appId: string, path: string, request: IApiRequest): Promise<IApiResponse> {
+        const api = this.providedApis.get(appId).get(path);
+
+        if (!api) {
+            return;
```
graywolf336: I think we should be returning a response here, even if it's just a 400 bad request. That way the bridged system doesn't have to handle nothing being returned and they actually get back a response when they call this method.

src/server/managers/AppApiManager.ts
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        const app = this.manager.getOneById(appId);
+
+        if (!app || AppStatusUtils.isDisabled(app.getStatus())) {
+            // Just in case someone decides to do something they shouldn't
+            // let's ensure the app actually exists
+            return;
```
graywolf336: Same here as my comment above

src/definition/api/IApi.ts
rodrigok: This already works 😄

src/definition/accessors/IApiExtend.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * This accessor provides methods for adding a custom api.
+ * It is provided during the initialization of your App
+ */
+
+export interface IApiExtend {
+    /**
+     * Adds an api which can be called by external services lateron.
+     * Should an api already exists an error will be thrown.
```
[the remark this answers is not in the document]
rodrigok: Idk if I understood your point, but you are able to provide more than one API, mainly if you need different properties like visibility and security. Is that what you were meaning?

src/definition/api/IApi.ts
rodrigok: I see your points, would it be better to have a class in place of the interface or a class that implements the interface named `ApiEndpoint`?

src/definition/api/IEndpoint.ts
```diff
@@ -0,0 +1,6 @@
+export interface IApiEndpointInfo {
```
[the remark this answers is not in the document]
rodrigok: 👍

src/definition/api/IApi.ts
```diff
@@ -0,0 +1,82 @@
...
+     * No security check will be executed agains the calls made to this URL
+     */
+    UNSECURE,
+    /**
+     * Only calls containing a valid token will be able to execute the api
+     * Mutiple tokens can be generated to access the api, by default one
+     * will be generated automatically.
+     * @param `X-Auth-Token`
+     */
+    // AUTH_TOKEN,
```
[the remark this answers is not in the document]
rodrigok: Well, the AUTH_TOKEN is already documented there, the idea is to have options to handle authentication automatically for the apps. But it's not clear the options for me now.

src/definition/api/IApi.ts
graywolf336: Now that's a good question, probably doesn't matter too much. Thinking about it though, might be best to just have one so there is less source for confusion when an external developer tries to look into the code to see how to do things.

src/server/managers/AppApi.ts
```diff
@@ -0,0 +1,126 @@
...
+    public readonly appId: string;
+    public readonly hash?: string;
+    public readonly implementedMethods: Array<string>;
+
+    constructor(public app: ProxiedApp, public api: IApi, public endpoint: IApiEndpoint) {
+        this.appId = app.getID();
+
+        switch (this.api.visibility) {
+            case ApiVisibility.PUBLIC:
+                this.basePath = `/api/apps/public/${app.getID()}`;
```
graywolf336: @rodrigok what would you think about us making the `nameSlug` of the Apps to be unique and use it instead of the App's ID? This would help with the transition over to #56 (package.json) to slowly get rid of the usage of the id. As a result, and where the idea came to me from, the "endpoints" and urls generated/provided will be easier to use (and shorter) instead of having a long id in the url.

src/definition/api/IApi.ts
graywolf336: @rodrigok sweet! Maybe you can add some documentation so other developers are aware of it and can make usage of it 😃

src/definition/api/IApi.ts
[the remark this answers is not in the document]
graywolf336: It's commented out, so it won't show up in the generated docs if I'm not mistaken..yeah I'm not sure the best route to go yet without putting forth much effort in thoughts.

src/definition/accessors/IApiExtend.ts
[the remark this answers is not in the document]
rodrigok: By the API we mean an endpoint, you can configure more than one API but containing different endpoints

src/definition/api/IApi.ts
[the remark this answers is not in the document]
rodrigok: I don't want to have it in docs since we don't have it implemented yet, it's basically a reminder that we will have more options there.

src/server/managers/AppApi.ts
rodrigok: I wan't to move the ID to be a unique ID based on name as NPM does, that way we keep using the ID here, just change how the ID would be generated.

src/definition/accessors/IApiExtend.ts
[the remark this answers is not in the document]
graywolf336: So, you can have more than one API but different endpoints? If I call this method more than once, each time containing two different endpoints, will it error out or will it succeed?

src/server/managers/AppApi.ts
graywolf336: Right, so if we go with the name slug or a special smaller ID such as `author:app-name` then we could generate shorter urls and that would be a unique ID since the author couldn't have two of the same app names but two people could have the same App name as their author name would be different.

src/definition/accessors/IApiExtend.ts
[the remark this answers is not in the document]
rodrigok: Will succeed

src/server/managers/AppApi.ts
rodrigok: Yes, but that will be in the future

meomay503:
HI @graywolf336 @rodrigok, does your team have a plan to implement the other security level for the API in the next releases?

rodrigok:
@meomay503, unfortunately, no, it's not in our short roadmap.

PR #74, opened 2018-10-22
[NEW] Action button types

graywolf336:
As the message actions require the type property and currently there is only one type of action item.


rodrigok:
@graywolf336 is this ok to be merged?

graywolf336:
@rodrigok yeah. Documentation is needed and I want to make it easier for App developers, but I can do that in another pull request.

PR #87, opened 2018-11-02
[FIX] Slash Command Contexts Not Supporting Usernames on Rooms

graywolf336:
The usernames property of the rooms on the slash command contexts was not updated to the new style of code, classes, which allows getting the usernames of the users in the rooms.

### DIFF
--- a/src/server/managers/AppSlashCommandManager.ts
+++ b/src/server/managers/AppSlashCommandManager.ts
@@ -320,19 +320,8 @@ export class AppSlashCommandManager {
             return;
         }
 
-        // Due to the internal changes for the usernames property, we need to ensure the room
-        // is a class and not just an interface
-        let room: Room;
-        if (context.getRoom() instanceof Room) {
-            room = context.getRoom() as Room;
-        } else {
-            room = new Room(context.getRoom(), this.manager);
-        }
-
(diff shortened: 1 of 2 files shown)


src/server/managers/AppSlashCommandManager.ts
```diff
@@ -319,8 +320,19 @@ export class AppSlashCommandManager {
...
+            room = context.getRoom() as Room;
+        } else {
+            room = new Room(context.getRoom(), this.manager);
+        }
+
+        const newContext = new SlashCommandContext(context.getSender(), room, context.getArguments());
+
         const appCmd = this.retrieveCommandInfo(cmd, app.getID());
-        await appCmd.runExecutorOrPreviewer(AppMethod._COMMAND_EXECUTOR, context, this.manager.getLogStorage(), this.accessors);
+        await appCmd.runExecutorOrPreviewer(AppMethod._COMMAND_EXECUTOR, newContext, this.manager.getLogStorage(), this.accessors);
```
rodrigok: A private method to process this and return the context would be better to prevent code duplication.

Than you could just call as:
```
await appCmd.runExecutorOrPreviewer(AppMethod._COMMAND_EXECUTOR, this.ensureContext(context), this.manager.getLogStorage(), this.accessors);
```

PR #89, opened 2018-11-05
Scheduler API

graywolf336:
To Do list:
- [x] Review codebase and merge conflicts
- [x] Find a new cron parser (as the package `later` is no longer active)
- [ ] Add tests covering the Scheduler components
- [ ] Implement bridges on Rocket.Chat side


src/definition/scheduler/Schedule.ts
```diff
@@ -0,0 +1,36 @@
+import { Recurrence, ScheduleData } from 'later';
+import { parse } from 'later';
```
mrsimpson: What about i18n for later? I've only seen english. If this is the only option, a reusable GUI component is elementary

rodrigok:
@graywolf336 is this preventing the jib to run multiple times when running rocket.chat in a cluster?

rodrigok:
@graywolf336 do you think this needs to be defined only inside the configuration method? Would be nice to define cron jobs based on messages or slashcommands, I think this should be something in the app class context.

src/definition/accessors/IJobBuilder.ts
```diff
@@ -0,0 +1,23 @@
...
+import { IRead } from './IRead';
+
+export interface IJobBuilder {
+    setName(name: string): IJobBuilder;
+    getName(): string;
+    setDescription(description: string): IJobBuilder;
+    getDescription(): string;
+    setData(data: { [key: string]: any }): IJobBuilder;
+    getData(): { [key: string]: any };
+    parseTextSchedule(textToParse: string): IJobBuilder;
```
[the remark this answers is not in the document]
graywolf336: Hehe, I had this in it's own builder but switched as it was a little confusing to use.

src/definition/accessors/IJobBuilder.ts
```diff
@@ -0,0 +1,23 @@
...
+    parseCronSchedule(cronToParse: string, includesSeconds?: boolean): IJobBuilder;
+    getRecurrenceScheduleBuilder(): RecurrenceBuilder;
+    setSchedule(schedule: Schedule): IJobBuilder;
+    getSchedule(): Schedule;
+    setExecutor(executor: (job: IJob, read: IRead, modify: IModify, http: IHttp, persis: IPersistence) => Promise<IJobResult>): IJobBuilder;
```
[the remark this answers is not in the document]
graywolf336: The builder is one way of several which a job can be created. The other primary way is to implement the `IJob` interface and then handle everything yourself. So yes, it can have it's own entity as scheduling a job accepts the `IJob` interface.

src/definition/scheduler/Schedule.ts
```diff
@@ -0,0 +1,36 @@
...
+
+function isScheduleData(arg: any): arg is ScheduleData {
+    return arg.schedules !== undefined;
+}
+
+export class Schedule implements ScheduleData {
+    public static fromText(text: string): Schedule {
```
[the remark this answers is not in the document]
graywolf336: 1. So people can easily generate a schedule outside of the builder.
2. Makes the builder an all-in-one, trying to make things easy for developers.

src/definition/scheduler/Schedule.ts
graywolf336: You bring up a very very good point, one which I need to reevaluate. I created this pull request in order to get feedback, so thank you for bringing this to my attention! :)

src/definition/accessors/IJobBuilder.ts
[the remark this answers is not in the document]
mrsimpson: I like this fluent interface pattern for builders, but only if the entity operated on is the same. And a job ain't a schedule, thus this mix was more confusing to me.

package.json
```diff
@@ -65,6 +66,7 @@
   },
   "dependencies": {
     "adm-zip": "^0.4.9",
+    "later": "^1.2.0",
```
dbrekelmans: In the project's [readme](https://github.com/bunkat/later#this-project-is-no-longer-being-maintained-it-should-be-treated-as-sample-code-on-one-technique-to-calculate-future-instances-of-complex-schedules-if-you-are-looking-to-fire-a-task-every-1-minute-use-a-cron-scheduler-instead):
> This project is no longer being maintained.

Might be an issue ;)

src/definition/accessors/ISchedulerExtend.ts
```diff
@@ -0,0 +1,8 @@
+import { IJob } from '../scheduler';
+import { IJobBuilder } from './IJobBuilder';
+
+export interface ISchedulerExtend {
```
dbrekelmans: I think this interface is missing some features.

You want to be able to unschedule jobs and get a list of scheduled jobs

src/definition/scheduler/IJob.ts
```diff
@@ -0,0 +1,18 @@
+import { IHttp, IModify, IPersistence, IRead } from './../accessors';
+import { IJobResult } from './IJobResult';
+import { Schedule } from './Schedule';
+
+export interface IJob {
```
dbrekelmans: Maybe you would want to store which user scheduled the job so admins can see who did what.
Or do you think that that's more an optional thing to store in `data`?

package.json
graywolf336: Indeed it is 🤔 now to find an alternative

src/definition/scheduler/IJob.ts
graywolf336: Yes sir. As you wouldn't want to limit the system to requiring a user since an App could provide "daily check ins" and that would need to be automatically scheduled when the app is installed without requiring a user to schedule it.

dbrekelmans:
@graywolf336 Is there anything I can do to help you work on this feature? I'm not very familiar with the apps-engine (only played around a bit with creating an app), but I'm willing to dedicate some time to help you work on this.

d-gubert:
@dbrekelmans hey! nice to know you're interested in helping us with this!

We're bringing this back up in our queue and will be dedicating time to the feature in the following weeks, and would definitely appreciate the help! We've updated the description to reflect what is still missing to get this published :)

src/definition/accessors/ISchedulerExtend.ts
graywolf336: So, what I'm going to do is add a new accessor to to the modify stage to allow modification to the jobs. But if I get into it and that is not a fluid design or works in reality, then we will see how else we can work with this.

src/definition/scheduler/IJob.ts
graywolf336: Added a `creator: IUser` property to the job and removed the `data` property since Apps should be using the persistence layer to store data.

src/definition/scheduler/IJob.ts
```diff
@@ -0,0 +1,23 @@
...
+import { IJobResult } from './IJobResult';
+
+export interface IJob {
+    /**
+     * The name of the job. As this property is not unique,
+     * we recommend setting this to something that is
+     * user friendly since the list of job names will be visible
+     * to the server administrator.
+     */
+    name: string;
```
graywolf336: Thoughts to everyone watching this thread about requiring an `id: string` property? As right now the `name: string` is simply for displaying on the UI the list of jobs and a history of them. But if an App wants to cancel a job or view the job's history there is not a way to do it currently.

src/definition/scheduler/Schedule.ts
```diff
@@ -0,0 +1,35 @@
+import { parseExpression } from 'cron-parser';
```
graywolf336: https://www.npmjs.com/package/cron-parser

src/definition/accessors/IJobBuilder.ts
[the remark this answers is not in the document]
graywolf336: I went ahead and did away with the builder. After looking into it, I didn't like it and felt it wasn't the right tool for this job.

src/definition/scheduler/Schedule.ts
graywolf336: As of right now, we will be switching to crontab style only. Support for i18n will come in the future.

src/definition/scheduler/IJob.ts
d-gubert: Couldn't we use the name for that? Prepend it with the `appId` and a colon and should be good to go IMO 😛

graywolf336: For our internal usage, yes. But for Apps to query, no. Because what if they want the same name for a job that runs based upon something a user does but doesn't want that detail to show up in the job history and they want to query all of the jobs done by a user? That's why I'm considering an id..

dbrekelmans:
Hey @d-gubert, I'm still interested in helping, but I'd need something concrete to work on. Maybe we can find something tomorrow if there's a place where we can chat.

I'm definitely going to write a remind-me app once this API is available. If it's up to standards maybe it could serve as an example for people looking to use this API.

src/definition/scheduler/IJob.ts
dbrekelmans: > Because what if they want the same name for a job that runs based upon something a user does but doesn't want that detail to show up in the job history and they want to query all of the jobs done by a user?

I'm not sure I quite understand what you mean by this. Could you maybe give an example?

graywolf336:
@dbrekelmans what's your username on the oepn.rocket.chat server?

src/definition/scheduler/IJob.ts
graywolf336: Let's say you only want one job to run but you have two internal jobs and their behavior is massively based upon a configuration. The setting is changed in the admin area and when that happens you want to replace the job, that way the admin isn't aware of the different types and only sees one job in the job history and that type of stuff. Does that make sense?

dbrekelmans:
@graywolf336 it's `dbrekelmans`, same as my github name.

src/definition/scheduler/IJob.ts
dbrekelmans: I don't understand why there would be one job running multiple internal jobs. Shouldn't that just be separate jobs then?

graywolf336: Imagine a payroll or something. For auditing purposes there should only appear as one job in the log. But because payroll is an extremely complicated item, the developer wants to separate the logic out from one giant job into different ones. So internally there are several jobs by the same name that way it appears as one job. I know this is an example that is stretching it, but I do see some valid reasons for not having the name be unique.

d-gubert: gotta admit I didn't follow the use case either :/

dbrekelmans: I think I understand what @graywolf336 means.
IMO if an app wants to split one scheduled job into multiple parts, then that's something the app itself can solve and not something that should be solved by the scheduling API

graywolf336: And that's why I'm think we probably should have an id for the jobs and not rely on the name 🤔

sveatlo:
Any progress on this? I need it desperately. The lack of a scheduler API is the last thing blocking me from completing a daily standup application for Rocket.Chat (like standup-raven for Mattermost), which in turn is the last bit I need to migrate the entire company to Rocket.Chat.
I am also willing to help with this merge request, but I'd need quite specific instructions about what to do, since my codebase knowledge is pretty much none at this point (I only tried creating like 2 rather simple applications). Let me know if I could help you.

graywolf336:
@sveatlo can you try out the current code base and get a feel for deploying an App using the current API? Then provide us feedback on it and ways we can improve?

majklovec:
Any progress?

meomay503:
I am looking forward to it :D

graywolf336:
Closing this out dated PR in favor of the new functionality that has been in the Apps Engine for a while now. :)

PR #91, opened 2018-11-09
[NEW] Create an AppAccessor for the Apps instances

d-gubert:
### Motivation

Currently, there is no way for an App to have access to the list of the Api Endpoints it created, nor is it possible to have access to any accessor at any given time.

I'm proposing in this PR an AppAccessor that does that. It is passed to the App at instantiation time, and makes it possible for the app instance to determine, at any given time, the API Endpoints it created (including the whole computed path of the endpoint), the SlashCommands it created, and any read accessors available, so there is a centralized way to access them in the methods the Apps Engine calls during its lifecycle.

This approach also enables us, Apps Engine developers, to add accessor to only one place, whenever it is created, not having to decide what method of the app gets access to it, and not having to pass an instance of it to every method that should access it.

Feedback is much welcome :)

### DIFF
--- a/src/definition/App.ts
+++ b/src/definition/App.ts
@@ -22,7 +22,7 @@ export abstract class App implements IApp {
      * Also, please use the `initialize()` method to do items instead of the constructor as the constructor
      * *might* be called more than once but the `initialize()` will only be called once.
      */
-    protected constructor(private readonly info: IAppInfo, private readonly logger: ILogger, private readonly accessors?: IAppAccessors) {
+    protected constructor(private readonly info: IAppInfo, private readonly logger: ILogger, public readonly accessors?: IAppAccessors) {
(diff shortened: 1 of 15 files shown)


d-gubert:
Should the `AppAccessors` class also provide access to instances of `ConfigurationExtend` or `ConfigurationModify`?

I think that maybe not `ConfigurationExtend`, maybe this access should be constrained to the `extendConfiguration` method as it is currently, or it could get confusing. And I'm not sure if the engine could handle a call to `configuration.api.provideApi` outside of that method. Being able to extend http configuration would be nice, though.

But I think it makes sense for an app to be able to alter server settings, for instance, at any given time, so the `ConfigurationModify` is a yes for me

rodrigok:
@d-gubert the options to modify the configurations can be added latter, by now it's better to make the access easy just to the accessors and provided configurations.

PR #92, opened 2018-11-16
[IMPROVE] Expose console for apps

rodrigok:


graywolf336:
No no no. Please revert.

This encourages people to use `console` and then will expect to see their logs inside of the App's "View Logs" and it won't be possible.

rodrigok:
@graywolf336 when we have a good way to see the logs on the webview we can pipe this to the logger, meanwhile it has been really hard to develop without it keeping refreshing the webview and expanding items to see the logs.

PR #117, opened 2019-03-28
Add msg_processing_type field to message action

d-gubert:
Adds the field introduced in [this PR](https://github.com/RocketChat/Rocket.Chat/pull/12626) to the engine.

No further changes are needed on Rocket.Chat for it to work.

### DIFF
--- a/src/definition/messages/MessageProcessingType.ts
+++ b/src/definition/messages/MessageProcessingType.ts
@@ -1,0 +1,4 @@
+export enum MessageProcessingType {
+    SendMessage = 'sendMessage',
+    RespondWithMessage = 'respondWithMessage',
+}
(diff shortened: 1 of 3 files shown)


src/definition/messages/MessageProcessingType.ts
```diff
@@ -0,0 +1,4 @@
+export enum MessageProcessingType {
+    SendMessage = 'sendMessage',
+    RespondWithMessage = 'respondWithMessage',
```
graywolf336: Wish there was documentation added that said what each of these do. They both seem the same to me :(

Does `sendMessage` actually send a message without populating the text area for the user to edit beforehand? Or is that `respondWithMessage` or what?

PR #118, opened 2019-04-03
Add setTimeout to vm context

marceloschmidt:
@d-gubert @rodrigok @graywolf336 can we add setTimeout to vm context? 
What are the bad implications in doing so? 
This would greatly help me in developing a new app, and I wouldn't have to wait until the scheduler is ready :)

### DIFF
--- a/src/server/compiler/AppCompiler.ts
+++ b/src/server/compiler/AppCompiler.ts
@@ -302,7 +302,7 @@ export class AppCompiler {
         }
 
         const customRequire = Utilities.buildCustomRequire(files);
-        const context = vm.createContext({ require: customRequire, exports, process: {}, console });
+        const context = vm.createContext({ require: customRequire, exports, process: {}, console, setTimeout });
 
         const script = new vm.Script(files[path.normalize(storage.info.classFile)].compiled);
         const result = script.runInContext(context);
--- a/src/server/compiler/AppCompiler.ts
+++ b/src/server/compiler/AppCompiler.ts
@@ -302,7 +302,7 @@ export class AppCompiler {
         }
 
         const customRequire = Utilities.buildCustomRequire(files);
-        const context = vm.createContext({ require: customRequire, exports, process: {}, console, setTimeout });
+        const context = vm.createContext({ require: customRequire, exports, process: {}, console, setTimeout, clearTimeout });
 
         const script = new vm.Script(files[path.normalize(storage.info.classFile)].compiled);
         const result = script.runInContext(context);


rodrigok:
I need to do some tests, but I think it may be possible.
What is the idea of usage?

d-gubert:
I think the "bad implications" would be the "raw" API it provides; you would have to calculate the milliseconds you want to wait/set the interval (as opposed to using a cron definition string), no "job" metadata/metrics, no retry on fail, timeouts do not persist on server restart, stuff like that.

Other than bad dev experience, not sure if there is technical problems on exposing it

rodrigok:
@d-gubert I see a usage for timeout out in cases like trying to simulate an human behavior sending the typing first and the text a few seconds later, for this kind of control I agree the timeout may be useful, just wondering if we should wrap the API and add a limit for it it the VM timeout doesn't cut the execution if it pass the time.

d-gubert:
> @d-gubert I see a usage for timeout out in cases like trying to simulate an human behavior sending the typing first and the text a few seconds later, for this kind of control I agree the timeout may be useful, just wondering if we should wrap the API and add a limit for it it the VM timeout doesn't cut the execution if it pass the time.

Yes, I think the use cases are valid, my concerns are HA setups (as mentioned above) and the VM timeout you've mentioned. I think wrapping the API is the way to go

marceloschmidt:
Uhmm... right, I didn't think about HA setups :(

marceloschmidt:
As for the files src/server/misc/Utilities.ts and src/server/ProxiedApp.ts they don't include the `console`, why should they include the `setTimeout`?

marceloschmidt:
Talking to @d-gubert I think HA wouldn't be affected by it, as it could be run inside a slashcommand executor, which would make it so it runs on a single instance. The problem is if the instance locks or is restarted, then anything on a timeout would be lost... IMO, that's an acceptable behavior.

marceloschmidt:
My use case is that I'm creating a `/send-later` command, that schedules a message to be sent at a given date/time.

graywolf336:
> The problem is if the instance locks or is restarted, then anything on a timeout would be lost... IMO, that's an acceptable behavior.

In my opinion, I don't think that is acceptable. A lot of these issues would be solved by the implementation of the scheduler work I started..I personally would rather see a concrete solution instead of a half-way done one.

Now, if we add a big warning that says "only use this for short time frames" then that's fine but we should not have people expect it to work across hours much less days. Also, if you do an http call and it times out, how does that affect the error handling and how does it handle logging?

d-gubert:
> Also, if you do an http call and it times out, how does that affect the error handling and how does it handle logging?

That's a good point... asynchronous things lose the ability to use the logger because, for the engine, the execution of the vm.context has already finished, and it has already closed the log session and saved it to the db.

graywolf336:
We will have the scheduler in by the end of November and with this PR sitting around for so long without any activity, I'm going to go ahead and close it in favor of the scheduler.

PR #125, opened 2019-06-10
Modifies the version regex to reflect marketplace's requirements

graywolf336:
Uncomplicates the version regex in the `app-schema.json` file. There might be other places where the version regex is applied, but I wanted to get this one in there since it was my reference for the marketplace.

### DIFF
--- a/src/definition/app-schema.json
+++ b/src/definition/app-schema.json
@@ -24,7 +24,7 @@
         "version": {
             "description": "The version of this App which will be used for display publicly and letting users know there is an update. This uses the semver format.",
             "type": "string",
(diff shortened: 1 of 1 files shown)


d-gubert:
Can you expand on what are the requirements for the marketplace?

One other thing, we are intending to release a `beta` (or `next`) version of the engine, will these RegExs allow that?

graywolf336:
The version of an app is only `number.number.number` and then the the engine requirements are the same but with the prefix of `~` or `^`. And no, the regexs won't allow that neither will the marketplace allow that.

d-gubert:
Ok.

Thinking about it, it makes sense - the marketplace should not have to accept apps expecting a `beta`/`next`, as shouldn't we as "App Reviewers". 

But would this prevent an app requiring a `beta`/`next` from being installed in RC with the _Development Mode_ enabled?

graywolf336:
Yeah, it makes sense...and even if we did want to include support for that we would have to rewrite that complex query to be even more complex 🙈 

This pull request won't. The only place this file is used is for IDE's to suggest and "validate" the json if people have hooked it up. Which 99.999999% probably haven't (or so I'd guess) 😂

graywolf336:
In the future though, we might pull this file and use it to validate an App's `app.json` file to ensure it is valid and contains what we want it to contain.

PR #126, opened 2019-07-01
Licensing strategy

d-gubert:


src/server/AppManager.ts
```diff
@@ -373,19 +377,33 @@ export class AppManager {
...
+                compiled: result.compiledFiles,
+                languageContent: result.languageContent,
+                settings: {},
+                implemented: result.implemented.getValues(),
+                marketplaceInfo,
+                status,
+            });
+
+            if (!created) {
+                aff.setStorageError('Failed to create the App, the satorage did not return it.');
```
graywolf336: `satorage` -> `storage`

src/server/marketplace/license/Crypto.ts
```diff
@@ -0,0 +1,24 @@
+import { publicDecrypt } from 'crypto';
+import { IInternalBridge } from '../../bridges';
+export class Crypto {
```
d-gubert:
```suggestion

export class Crypto {
```

src/server/managers/AppLicenseManager.ts
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        if (renewal < now) {
+            validationResult.addWarning('renewal', 'License has expired and needs to be renewed');
+        }
+
+        if (license.seats < currentActiveUsers) {
+            validationResult.addWarning(
+                'seats',
+                'License does not accomodate the currently active users. Please expand number of seats',
```
graywolf336: `Your license for the ${ app.name } does not have enough seats to accommodate the current amount of active users. Please increase the number of seats.`

How's that sound?

src/server/managers/AppLicenseManager.ts
```diff
@@ -0,0 +1,79 @@
...
+        try {
+            license = await this.crypto.decryptLicense(subscriptionInfo.license.license) as any;
+        } catch (err) {
+            validationResult.addError('publicKey', err.message);
+
+            throw new InvalidLicenseError(validationResult);
+        }
+
+        if (license.appId !== appId) {
+            validationResult.addError('appId', `License hasn't been issued for this app`);
```
graywolf336: I wonder about using the context phrase "this" when using error codes. Guess it depends on how the i18n will work for these and how the errors/warnings will be displayed.

src/server/managers/AppLicenseManager.ts
d-gubert: Much better :) Will apply

src/server/managers/AppLicenseManager.ts
d-gubert: Great point, certainly haven't taken that into account. Maybe the engine return error codes instead of messages, and then Rocket.Chat can apply i18n?

graywolf336: Possibly? Maybe return both? Better to be verbose than not when it comes to trying to determine where a bug is..

d-gubert: Well, then I guess it's ok - Rocket.Chat will be able to differentiate the errors with the `field` property (`'appId'` in this case)

PR #129, opened 2019-08-02
createBot to the IUserRead

Cool-fire:
1. Implemented the createBot interface in IUserRead.

This is an  initial pr to show the proof of concept for creating bot users with apps.

### DIFF
--- a/src/definition/accessors/IUserRead.ts
+++ b/src/definition/accessors/IUserRead.ts
@@ -8,4 +8,6 @@ export interface IUserRead {
     getById(id: string): Promise<IUser>;
 
     getByUsername(username: string): Promise<IUser>;
+
+    createBot(username: string, email: string, name: string, password: string): Promise<void>;
 }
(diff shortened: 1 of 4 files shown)


Cool-fire:
Thoughts
1. this interface should not have password field, instead a random password is generated automatically and needss to find a way to display it in the GUI. 
2. This interface should be  better if present in IModify instead  of IUserRead since we are creating a bot

src/definition/accessors/IUserRead.ts
```diff
@@ -8,4 +8,6 @@ export interface IUserRead {
     getById(id: string): Promise<IUser>;
 
     getByUsername(username: string): Promise<IUser>;
+
+    createBot(username: string, email: string, name: string, password: string): Promise<void>;
```
d-gubert: I think we would have to create another interface, maybe a `IUserCreator`, to define this method.

Also, you could actually receive as a parameter either a full `IUser` object or a subset of that, for instance `IBotUser`, containing only the properties that you need.

PR #132, opened 2019-08-08
Apps.engine.ui.sdk

lolimay:
Hi @d-gubert. I implemented AppEmbeddedSDK, can you please review it?

### DIFF
--- a/src/client/AppEmbeddedSDK.ts
+++ b/src/client/AppEmbeddedSDK.ts
@@ -19,10 +19,9 @@ export class AppEmbeddedSDK {
     }
 
     /**
-     * Create a message listener with id and add them to callbacks map set.
-     * @param id the id of the message event listener
+     * Initialize the App Embedded SDK for communicating with Rocket.Chat
      */
-    private initListener(): void {
+    public init(): void {
         this.listener = ({ data }) => {
             if (!data.hasOwnProperty('rcEmbeddedSDK')) {
                 return;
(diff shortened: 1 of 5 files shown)


src/client/AppEmbeddedSDK.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    public getRoomInfo(): Promise<any> {
+        return this.call(AppEmbeddedSDKActions.GET_ROOM_INFO);
+    }
+
+    /**
+     * Create a message listener with id and add them to callbacks map set.
+     * @param id the id of the message event listener
+     */
+    private initListener(): void {
```
d-gubert: I guess this method could be public, so that the developer of the `externalComponent` can choose when to initialise the Apps-Engine UI lib. Then you can remove the `this.initListener` call from the `call` method

src/client/AppEmbeddedSDK.ts
```diff
@@ -0,0 +1,63 @@
...
+        };
+        window.addEventListener('message', this.listener);
+    }
+    private call(action: string, payload?: any ): Promise<any>  {
+        return new Promise((resolve) => {
+            const id = randomString(ACTION_ID_LENFTH);
+
+            this.initListener();
+
+            if (window.parent) {
```
d-gubert: Funny thing is `window.parent` is always available - if the window has no actual parent, the `window.parent` property is only a reference to itself :) So you don't need to check this.

[As stated on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/parent)

src/client/AppEmbeddedSDK.ts
```diff
@@ -0,0 +1,63 @@
...
+                this.callbacks.delete(id);
+            }
+        };
+        window.addEventListener('message', this.listener);
+    }
+    private call(action: string, payload?: any ): Promise<any>  {
+        return new Promise((resolve) => {
+            const id = randomString(ACTION_ID_LENFTH);
+
+            this.initListener();
```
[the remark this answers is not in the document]
lolimay: Yes. Each API call will cause one listener to be initialized.

src/client/AppEmbeddedSDK.ts
lolimay: OK. I try it.

src/client/AppEmbeddedSDK.ts
lolimay: Thanks!

src/client/AppEmbeddedSDK.ts
lolimay: Thanks for your suggestion! I've make the initListener to be public.

PR #133, opened 2019-08-20
Interface to create Bot user through Apps.engine

Cool-fire:
### changes made:
- [x] Added a new Interface IUserGenerate for creating user
- [x] Added IUserBuilder to create a user 
- [x] Added a new Interface create in IUserBridge which accepts IUserGenerate

PR for implementation Rocket.Chat: [Link](https://github.com/RocketChat/Rocket.Chat/pull/15209)


Cool-fire:
@d-gubert Please review

Cool-fire:
sure @graywolf336 I will definitely add some test for user builder

graywolf336:
Thanks! I'll work with @d-gubert to see about getting this merged once you do.

graywolf336:
@Cool-fire let me know if you need help with anything

src/definition/users/IUserGenerate.ts
```diff
@@ -0,0 +1,12 @@
+export interface IUserGenerate {
```
graywolf336: I'm not 100% certain about this interface name. What is the main reason for not putting this on the existing user? What about extending the `IUser` interface so that it accepts everything else along with the new fields that are just for when creating a user?

src/server/accessors/ModifyCreator.ts
```diff
@@ -72,4 +85,10 @@ export class ModifyCreator implements IModifyCreator {
 
         return this.bridges.getRoomBridge().create(result, builder.getMembersToBeAddedUsernames(), this.appId);
     }
+
+    private _finishUser(builder: IUserBuilder): Promise<string> {
+        const result = builder.getUser();
+        delete result.id;
+        return this.bridges.getUserBridge().create(result, this.appId);
```
graywolf336: Almost wonder if we should do a check to ensure the user doesn't exist by the username, email, etc before we proceed to create the user that way we can gracefully handle the error instead of letting Rocket.Chat handle the error for us.

src/server/accessors/UserBuilder.ts
```diff
@@ -0,0 +1,119 @@
...
+            throw new Error('The "username" property is required.');
+        }
+        if (!this.user.email) {
+            throw new Error('The "email" property is required.');
+        }
+        if (!this.user.name) {
+            throw new Error('The "name" property is required.');
+        }
+        if (!this.user.roles) {
+            throw new Error('The "roles" property is required.');
```
graywolf336: We almost need a way for Apps to get fetch the roles from the Rocket.Chat system 🤔

src/definition/accessors/IModify.ts
```diff
@@ -554,3 +556,131 @@ export interface IRoomBuilder {
...
+    */
+    setDisplayName(name: string): IUserBuilder;
+
+   /**
+    * Gets the display name of this user.
+    */
+    getDisplayName(): string;
+
+    /**
+     * Sets the username for the user
```
graywolf336: Might want to add a warning here that the username must be **unique**.

src/definition/accessors/IModify.ts
```diff
@@ -554,3 +556,131 @@ export interface IRoomBuilder {
...
+   * @param user the user data to set
+   */
+   setData(user: IUserGenerate): IUserBuilder;
+
+   /**
+    * Sets the email of the user
+    *
+    * @param email the email address of the user
+    */
+    setEmail(email: string): IUserBuilder;
```
graywolf336: The email provided here can not exist already in the system otherwise it will cause an error to arise.

src/definition/users/IUserGenerate.ts
Cool-fire: @graywolf336 The main reason for not extending the `IUser` is that there are some fields which are non-nullable like `lastLoginAt`, `createdAt` which I felt are to be extended unnecessarily while creating, So I have created a new Interface which is meant for the only generation of a user.

Cool-fire: sure @graywolf336 I will change that as you suggested :)

graywolf336: Awesome, thank you! 😄

src/server/accessors/ModifyCreator.ts
Cool-fire: Yes, @graywolf336 it would be better if we can check a user exists by username, email..etc but I couldn't find a way to check whether it exists from apps-engine. Anyways I think the server handles the case where a user exists gracefully throwing the reason for the error in logs.

wreiske:
Would it be possible to add the ability for an app to edit a user, as well as create it? That would be amazing! Please, and thank you :)

src/server/accessors/ModifyCreator.ts
graywolf336: Then let's create it that way we can throw custom errors. The Apps Engine is an abstraction away from Rocket.Chat since the internals of Rocket.Chat can change at any point in time and we want the Apps Engine to be stable and consistent. Thus I would rather us do the extra work that way it is not a guessing game for App developers of what will happen and instead they can know up front. Does that make sense?

Cool-fire:
@graywolf336 can you review the PR, I think I have added necessary tests for UserBuilder. Please let me know if any changes are needed.

graywolf336:
@Cool-fire will do, probably will be tomorrow or Thursday before I can.

Based upon your experience with this, how hard would it be to implement what @wreiske requested?

Cool-fire:
@graywolf336 I think this pr gives the apps ability to create users. And I think it would be simple to make an interface like **IUserModifier** which has the ability to edit a user details but I think we have to look take a look on permissions to edit a user roles since I think that can be only done by an admin or the same user, I am wondering if the app developers have to use an admin role user to edit other user details.

d-gubert:
We've merged part of this implementation on https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/195, along with its Rocket.Chat counterpart.

Thanks again @Cool-fire ! 🤗

PR #137, opened 2019-09-10
Add new location to the external components called MODAL

lolimay:
image

### DIFF
--- a/src/definition/externalComponent/IExternalComponent.ts
+++ b/src/definition/externalComponent/IExternalComponent.ts
@@ -32,7 +32,7 @@ export interface IExternalComponent {
 }
 
 export enum ExternalComponentLocation {
-    CONTEXTUAL_BAR,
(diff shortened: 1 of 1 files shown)


d-gubert:
This needs implementation on the Rocket.Chat side to properly work, right? 🤔

lolimay:
> This needs implementation on the Rocket.Chat side to properly work, right?

Yes, we also need implementations in the Rocket.Chat side

src/definition/externalComponent/IExternalComponent.ts
```diff
@@ -32,5 +32,7 @@ export interface IExternalComponent {
 }
 
 export enum ExternalComponentLocation {
-    CONTEXTUAL_BAR,
+    CONTEXTUAL_BAR = 'CONTEXTUAL_BAR',
+
```
d-gubert:
```suggestion
```

PR #138, opened 2019-09-16
Deploy versions to NPM via CI

rodrigok:
@d-gubert need to define if the publish process should increase the version and set the build number.


d-gubert:
@rodrigok do you intend to make more changes to this PR or can it be merged?

rodrigok:
@d-gubert If you agree with the implementation you can merge it

PR #145, opened 2019-09-27
Alpha (1.9.0)

d-gubert:
Changes:

* [NEW] External Components


graywolf336:
Can we get a better pull request name?

src/client/AppEmbeddedSDK.ts
```diff
@@ -0,0 +1,57 @@
...
+export class AppEmbeddedSDK {
+    private listener: (this: Window, ev: MessageEvent) => any;
+    private callbacks: Map<string, (response: any) => any>;
+
+    constructor() {
+        this.listener = () => console.log('init');
+        this.callbacks = new Map();
+    }
+
+    public getUserInfo(): Promise<any> {
```
graywolf336: `Promise<any>`?

src/client/AppEmbeddedSDK.ts
```diff
@@ -0,0 +1,57 @@
...
+        return this.call(AppEmbeddedSDKActions.GET_USER_INFO);
+    }
+
+    public getRoomInfo(): Promise<any> {
```
graywolf336: `Promise<any>`? Also, where is the documentation for this?

src/definition/externalComponent/IWebHooks.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * This interface provides webhooks to the game. You can set
```
graywolf336: Not sure I understand why we will have a second webhook/endpoint in the codebase. As a developer, how will I know when to use what? Why can't i use the other endpoint system already in the apps for these external components? Also, this documentation still refers to games.

src/definition/externalComponent/IWebHooks.ts
```diff
@@ -0,0 +1,10 @@
...
+ * the target URL for each webhook event according to your demand.
+ */
+export interface IWebHooks {
+    // When a game is closed, Rocket.Chat will post all
```
graywolf336: ~~`game`~~

src/definition/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rocket.chat/apps-ts-definition",
-  "version": "1.5.2",
+  "version": "1.6.0-alpha",
```
graywolf336: Two different versions in the various packages.

src/server/managers/AppExternalComponentManager.ts
```diff
@@ -0,0 +1,74 @@
...
+    public unregisterComponent(appId: string): void {
+        if (this.providedComponents.has(appId)) {
+            this.providedComponents.delete(appId);
+        }
+    }
+    /**
+     * Get the external component by the appId.
+     * @param appId the id of the app
+     */
+    public getExternalComponent(appId: string): IExternalComponent {
```
graywolf336: I'm confused. Based upon the naming convention of nearly all of the methods in this class, an app can only have one external component. Yet the variable names inside here say otherwise. Can we update the method names to reflect what is happening? And if it is the case that only one external component is allowed per app, then we need to allow more than one per App before this gets merged. Unless there is a bigger technical limitation as to why we are limited to only one per app.

tests/test-data/bridges/externalComponentBridge.ts
```diff
@@ -0,0 +1,8 @@
+import { IExternalComponent } from '../../../src/definition/externalComponent';
+import { IExternalComponentBridge } from '../../../src/server/bridges';
+export class TestExternalComponentBridge implements IExternalComponentBridge {
+    // TODO Need to add the actual tests here
```
graywolf336: Looking forward to seeing these tests. :)

tests/test-data/bridges/appBridges.ts
```diff
@@ -28,6 +28,7 @@ import { TestsUserBridge } from './userBridge';
 export class TestsAppBridges extends AppBridges {
     private readonly appDetails: TestsAppDetailChangesBridge;
     private readonly cmdBridge: TestsCommandBridge;
+
```
graywolf336:
```suggestion

```

src/server/managers/AppExternalComponentManager.ts
```diff
@@ -0,0 +1,74 @@
...
+        if (this.appTouchedComponents.has(appId)) {
+            return this.appTouchedComponents.get(appId);
+        }
+
+        return null;
+    }
+    /**
+     * Remove the external compoent from both the providedComponents
```
graywolf336:
```suggestion
     * Remove the external component from both the providedComponents
```

src/server/managers/AppExternalComponentManager.ts
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public registerComponent(appId: string, externalComponent: IExternalComponent, addToProvidedComponents: boolean = false): void {
+        this.appTouchedComponents.set(appId, externalComponent);
+
+        if (addToProvidedComponents) {
+            this.providedComponents.set(appId, externalComponent);
+        }
+    }
+    /**
+     * Remove the external compoent from the providedComponents by appId.
```
graywolf336:
```suggestion
     * Remove the external component from the providedComponents by appId.
     * 
```

src/server/managers/AppExternalComponentManager.ts
```diff
@@ -0,0 +1,74 @@
...
+     * or enabled. As long as an app touched an external component, then
+     * it is listed here.
+     */
+    private appTouchedComponents: Map<string, IExternalComponent>;
+    constructor() {
+        this.providedComponents = new Map<string, IExternalComponent>();
+        this.appTouchedComponents = new Map<string, IExternalComponent>();
+    }
+    /**
+     * Add the external component to the appTouchedComponents.
```
graywolf336:
```suggestion
     * Add the external component to the appTouchedComponents.
     * 
```

src/server/managers/AppExternalComponentManager.ts
```diff
@@ -0,0 +1,74 @@
...
+    private providedComponents: Map<string, IExternalComponent>;
+    /**
+     * Contains the apps and the external components they have touhed.
+     * The key is appId and the value is external component.
+     * Doesn't matter whether the app provided, modified, disabled,
+     * or enabled. As long as an app touched an external component, then
+     * it is listed here.
+     */
+    private appTouchedComponents: Map<string, IExternalComponent>;
+    constructor() {
```
graywolf336:
```suggestion

    constructor() {
```

src/server/managers/AppExternalComponentManager.ts
```diff
@@ -0,0 +1,74 @@
+import { IExternalComponent } from '../../definition/externalComponent';
+
+/**
+ * The external component manager for the Apps.
+ *
+ * An App will register external component during their `initialize` method.
```
graywolf336:
```suggestion
 * An App will register an external component during their `initialize` method.
```

src/server/accessors/ExternalComponentsExtend.ts
```diff
@@ -0,0 +1,12 @@
+import { AppExternalComponentManager } from '../managers/AppExternalComponentManager';
+
+import { IExternalComponentsExtend } from '../../definition/accessors';
+import { IExternalComponent } from '../../definition/externalComponent/IExternalComponent';
+
+export class ExternalComponentsExtend implements IExternalComponentsExtend {
+    constructor(private readonly manager: AppExternalComponentManager, private readonly appId: string) { }
+
+    public async register(externalComponent: IExternalComponent): Promise<void> {
```
graywolf336: What will happen when an App calls register multiple times with different `IExternalComponent` instances? Based upon how the precedence of the slash commands, it _should_ add the different instances and have all of them being valid. However, looking elsewhere in this external component code it does not behave the same way. We need the behavior/logic to be consistent if we are using the same terminology within the codebase.

src/definition/accessors/IExternalComponentsExtend.ts
```diff
@@ -0,0 +1,9 @@
+import { IExternalComponent } from '../externalComponent';
+
+/**
+ * This accessor provides methods for register an external
+ * component. Provided during the initialization of your App.
+ */
+export interface IExternalComponentsExtend {
+    register(externalComponent: IExternalComponent): Promise<void>;
```
graywolf336: Please add documentation to this method and see my other comment on this class's implementation.

src/definition/accessors/IConfigurationExtend.ts
```diff
@@ -17,4 +18,6 @@ export interface IConfigurationExtend {
     /** Accessor for declaring the commands which your App provides. */
     readonly slashCommands: ISlashCommandsExtend;
     readonly api: IApiExtend;
+
+    readonly externalComponents: IExternalComponentsExtend;
```
graywolf336: Another place where the terminology is consistent with the codebase yet the logic is not consistent. Plural is staying consistent yet the logic of being singular is not consistent since our other logic, for slash commands, is also plural.

src/client/AppEmbeddedSDK.ts
lolimay: Hi @graywolf336 . Here is the temporal documentation for the AppEmbeddedSDK: https://lolimay.gitbook.io/workspace. Can you please also review it? It will be great if you can also leave some comments :)
I'll create a PR to add this document to the docs repository if there's no much problem.

graywolf336: It looks okay for external documentation, however, our current codebase has some documentation generated here: https://rocketchat.github.io/Rocket.Chat.Apps-engine/. Thus it would be not only beneficial but also consistent to have documentation inside the codebase on the code. Another thing, please remove the `any` and have it be typed so that developers know exactly what to expect and they get code suggestions in their IDE which is one of the biggest benefits for using TypeScript.

lolimay: Got it. Thanks! I will do it.

graywolf336: Also, I see that in your documentation people have to link to a CDN for the file. We have a large percentage of Rocket.Chat servers who operate offline and with their Rocket.Chat instance not being able to talk to the internet. How will external components work in this instance? We are working on enabling "air gapped" servers to download and install Rocket.Chat Apps from the Marketplace, which means Apps will need to be able to operate fully offline. How will external components work with these type systems?

lolimay: It's a temporal solution that the FRVR team can use our AppEmbeddedSDK for testing at an early stage. Later, we will release compiled SDK files on the npm. So that the developers can `npm install` to install the AppEmbeddedSDK and buddle it with their apps together.

And I don't think the external components can be loaded in an offline environment. Because the HTML5 game is usually served on a game company server or a CDN. This means the "air-gapped" hoster should give up the external component feature that can play chat-based HTML5 games in Rocket.Chat.

lolimay: @geekgonecrazy I made a [PR](https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/147) to add types for the public AppEmbeddedSDK methods and add documentations for them. Can you please review it?

graywolf336: Okay. So, it sounds like we might need to add a flag for app publishers to tell us which Apps support "air gapped" servers and which ones require an internet connection. @d-gubert and @geekgonecrazy would like your opinions and thoughts on this.

src/definition/package.json
lolimay: @d-gubert Do you know why it happened?

src/server/managers/AppExternalComponentManager.ts
lolimay: Hi @graywolf336. Sorry for the confusion brought to you.

> an app can only have one external component.
yeah, it's true. We only allow one external component for one app currently. It's much easier for a Rocket.Chat administrator to enable/disable an external component by just enabling the app or disabling/uninstalling an app. It will make the problem become more complex if we support that one app owns multiple external components (We will need to add more control options in the apps admin panel for administrators to control).

> yet the variable names inside here say otherwise. Can we update the method names to reflect what is happening? 
the class AppExternalComponentManager maintains all the external components from all the apps installed in the Rocket.Chat. As I said before - one app corresponding to one external component, so when an administrator enabled/disabled an app, we need to enable/disable its external component correspondingly (here we need the method getExternalComponent, which returns an external component by the appId).

src/definition/accessors/IConfigurationExtend.ts
lolimay: Sorry, according to the previous discussions (one app only owns one external component). I think here it should be `externalComponent: IExternalComponentExtend`.

tests/test-data/bridges/externalComponentBridge.ts
lolimay: We finally use AppExternalComponentManager to maintains all the external components itself. Not depends on the bridged system anymore. This file should be removed.

lolimay: resolved in https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/149/commits/6a522751da78a0ab521b1b1e34587ea0c513d5f7

src/server/managers/AppExternalComponentManager.ts
lolimay: resolved in https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/149/commits/2a4e5ca641b067de0dd4d609d7f1877478e863d3

tests/test-data/bridges/appBridges.ts
lolimay: resolved in https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/149/commits/2a4e5ca641b067de0dd4d609d7f1877478e863d3

src/server/managers/AppExternalComponentManager.ts
lolimay: resolved in https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/149/commits/ba0118e1832aceaad8285dd005ea4a05bdba4135

src/server/managers/AppExternalComponentManager.ts
lolimay: https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/149/commits/ba0118e1832aceaad8285dd005ea4a05bdba4135

src/server/managers/AppExternalComponentManager.ts
lolimay: resolved in https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/149/commits/ba0118e1832aceaad8285dd005ea4a05bdba4135

src/server/managers/AppExternalComponentManager.ts
lolimay: resolved in https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/149/commits/ba0118e1832aceaad8285dd005ea4a05bdba4135

src/server/managers/AppExternalComponentManager.ts
lolimay: resolved in https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/149/commits/b49dcf1ccc902d4c766e671a609cc2b4f4c31678

src/client/AppEmbeddedSDK.ts
geekgonecrazy: I agree I think would be worth while to identify all apps that must have an internet connection to work.  I think this would even be useful for displaying to admin's.

rodrigok: I agree, and that may be a permission in the future as well.

src/definition/accessors/IConfigurationExtend.ts
d-gubert: @lolimay is there a technical reason why we wouldn't allow an app to register more than one `externalComponent`? For me, considering the functionality, an app could register a MODAL component to show at some point in time and a CONTEXTUAL_BAR in another - it doesn't look like there should be a problem

src/server/managers/AppExternalComponentManager.ts
d-gubert: > It's much easier for a Rocket.Chat administrator to enable/disable an external component by just enabling the app or disabling/uninstalling an app. It will make the problem become more complex if we support that one app owns multiple external components (We will need to add more control options in the apps admin panel for administrators to control).

I'm not so sure about this. We may only allow one external component per app now if it is technically much more difficult to implement, but as I stated in another comment, I see valid use cases for an app registering more than one external component

src/definition/externalComponent/IWebHooks.ts
d-gubert: The idea is to replace this with event listener interfaces, similar to `IPreMessageSent`

src/server/managers/AppExternalComponentManager.ts
lolimay: OK. Will try to support multiple external components per app. @graywolf336 sorry for the confusion caused to you.

src/definition/externalComponent/IWebHooks.ts
graywolf336: Yes, that sounds like a much better solution. 👍

src/client/AppEmbeddedSDK.ts
graywolf336: As of right now we could assume that. If they use `IHttp` or implement this interface, they require an internet connection to work.

lolimay: An app that needs to register an external component also possibly requires an Internet connection to work (The external component usually creates an `<iframe>` container to load HTML5 game by the external URL).

d-gubert: Yes - but in a different moment. The "airgapped" we usually refer is regarding the servers, so if your app wants to make a request to the internet in a slashcommand, for instance, it wouldn't be possible.

But as the `<iframe>` is rendered on the UI, the user would more likely have access to the outer world - but maybe restricted by some firewall rules

src/server/managers/AppExternalComponentManager.ts
lolimay: resolved in the PR: https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/150

src/server/accessors/ExternalComponentsExtend.ts
graywolf336: Still wondering what happens when they call this method twice and provide different components. Will this error out or will it fail silently or will it work and if it works then which instance is valid?

lolimay: Hi @graywolf336 . The `register` method actually calls the method `addExternalComponnet` of the AppExternalComponent. 
- when we call this method twice and provide different components, the app will own 2 external components.
- when we call this method twice and provide the same components, the will still be an external component (we store external components in a map).
- when we call this method twice and provide the different components **but with the same name**, we will throw an ExternalComponentAlreadyTouchedError.
https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/adc01ab56f7cbd945ce395de24aa56b172fb6685/src/server/managers/AppExternalComponentManager.ts#L74-L80

---

Below is the latest implementation and unit tests:

https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/83d2ba18c859fe3eaf87d634ff1c2e3b640beb63/src/server/managers/AppExternalComponentManager.ts#L57-L81

https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/83d2ba18c859fe3eaf87d634ff1c2e3b640beb63/tests/server/managers/AppExternalComponentManager.spec.ts#L85-L110

src/server/managers/AppExternalComponentManager.ts
graywolf336: No confusion here, I just wanted to bring the discussion up to understand the reasoning behind it. And that’s what this pull request is for, to discuss and improve. 🙂

We will need to list the external components in the configuration page of the App inside Rocket.Chat. That way an administrator can easily enable or disable what they want. We will also need a way for an app to tell what the status of a component is.

src/definition/package.json
d-gubert: This is related to the `@rocket.chat/apps-ts-definition` package, for which we should drop updating. In fact, we haven't been updating it with the latest releases of the Engine.  We should remove that in some upcoming version, but I guess we could leave it there for now and deal with that in the beta branch

src/definition/accessors/IExternalComponentsExtend.ts
lolimay: resolved in the PR: https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/156

PR #147, opened 2019-10-01
Resolve reviews: typed public AppEmbeddedSDK methods and add documentations for them

lolimay:
Resolve: https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/145#discussion_r330047784
Resolve: https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/145#discussion_r330048038
Resolve: https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/145#discussion_r330142696

### DIFF
--- a/src/client/interfaces/IRoomInfo.ts
+++ b/src/client/interfaces/IRoomInfo.ts
@@ -1,0 +1,17 @@
+import { IUserInfo } from './IUserInfo';
+
+export interface IRoomInfo {
+    /**
+     * the id of the Rocket.Chat room
+     */
+    roomId: string;
+    /**
+     * the room name of the Rocket.Chat room
+     */
+    roomName: string;
+    /**
+     * the list that contains all the users belonging
+     * to this room.
+     */
+    members: Array<IUserInfo>;
+}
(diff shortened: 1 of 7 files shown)


src/client/interfaces/IRoomInfo.ts
```diff
@@ -0,0 +1,17 @@
+import { IUserInfo } from './IUserInfo';
+
+export interface IRoomInfo {
```
graywolf336: Let's call these something client side specific, otherwise developers working on the backend/server side could easily get confused as to which one to use.

src/client/interfaces/IUserInfo.ts
```diff
@@ -0,0 +1,14 @@
+export interface IUserInfo {
```
graywolf336: Let's call these something client side specific, otherwise developers working on the backend/server side could easily get confused as to which one to use.

src/client/interfaces/IRoomInfo.ts
lolimay: Great practice! Learned a lot 😃

src/client/interfaces/IUserInfo.ts
lolimay: changed

PR #150, opened 2019-10-03
Support multiple external components per app

lolimay:
@RocketChat/Rocket.Chat.Apps-engine


src/server/AppManager.ts
```diff
@@ -758,15 +758,15 @@ export class AppManager {
         }
 
         if (enable) {
-            const externalComponent = this.externalComponentManager.getExternalComponent(app.getID());
+            const externalComponents = this.externalComponentManager.getExternalComponents(app.getID());
 
             this.commandManager.registerCommands(app.getID());
-            this.externalComponentManager.registerComponent(app.getID(), externalComponent, true);
+            this.externalComponentManager.registerExternalComponents(app.getID(), externalComponents);
```
graywolf336: Why are you passing in the external components to register when the `externalComponentManager` already has those stored in its instance?

src/server/errors/ExternalComponentAlreadyTouchedError.ts
```diff
@@ -0,0 +1,10 @@
+import { IExternalComponent } from '../../definition/externalComponent';
+
+export class ExternalComponentAlreadyTouchedError implements Error {
+    public name: string = 'ExternalComponentAlreadyTouched';
+    public message: string;
+
+    constructor(externalComponent: IExternalComponent) {
+        this.message = `The app(${externalComponent.appId}) has already touched the external component(${externalComponent.name})`;
```
graywolf336:
```suggestion
        this.message = `The app (${externalComponent.appId}) has already touched the external component (${externalComponent.name})`;
```

src/server/managers/AppExternalComponentManager.ts
```diff
@@ -1,79 +1,129 @@
...
- * only then will that App's external component be enabled.
+ * An app will register external components during its `initialize` method.
+ * Then once an app's `onEnable` method is called and it returns true,
+ * only then will that app's external components be enabled.
  */
 export class AppExternalComponentManager {
     /**
-     * The map that maintain the list of all registered components.
+     * The map that maintains all registered components.
+     * The key of the top map is app id and the key of inner map is the
```
graywolf336: With this addition, you may want to modify the documentation on the `IExternalComponent` to state that the `name` must be unique.

src/server/AppManager.ts
lolimay: no @graywolf336 . before this executed, the `externalComponentManager` does not have any external components in `providedExternalComponents`. The `addExternalComponent` method only add the external component to `apTouchedExternalComponents`, and `registerExternalComponent` will actually add them to `providedExternalComponents` during an app enabling period.

src/server/errors/ExternalComponentAlreadyTouchedError.ts
lolimay: Thanks!

src/server/AppManager.ts
lolimay: image

src/server/errors/ExternalComponentAlreadyTouchedError.ts
lolimay: resolved in https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/150/commits/f64e05be3c49c00693a37cdcf862d77cc6e11149

d-gubert: Wonder if our lint shouldn't be hinting at this kind of problems 🤔 Will be checking

graywolf336: That would be great to have a linter for stuff like this.

src/server/managers/AppExternalComponentManager.ts
```diff
@@ -1,79 +1,129 @@
...
+        if (!this.appTouchedExternalComponents.get(appId)) {
+            this.appTouchedExternalComponents.set(appId, new Map(Object.entries({ [externalComponent.name]: externalComponent})));
+        } else {
+            const appExternalComponents = this.appTouchedExternalComponents.get(appId);
+            const touchedExternalComponent = appExternalComponents.get(externalComponent.name);
+
+            if (
+                touchedExternalComponent
+                && touchedExternalComponent.name === externalComponent.name
+                && JSON.stringify(touchedExternalComponent) !== JSON.stringify(externalComponent)
```
graywolf336: I'm not sure if I understand the need for this, can you explain? If we are making the `name` part the unique aspect, then why would we care about if they are different on the internals? I don't think we should care about it as with this check it makes this call limit what can be done. What if I want the same component name but want to change the content when they make a setting change and only need/want that one component? What I just said could also be an undesired usage we want, but I think it would be fine 🤔

src/server/managers/AppExternalComponentManager.ts
```diff
@@ -1,79 +1,129 @@
...
     /**
-     * The map that maintain the list of all registered components.
+     * The map that maintains all registered components.
+     * The key of the top map is app id and the key of inner map is the
+     * external component name.
      */
-    private providedComponents: Map<string, IExternalComponent>;
+    private providedExternalComponents: Map<string, Map<string, IExternalComponent>>;
     /**
      * Contains the apps and the external components they have touhed.
```
graywolf336:
```suggestion
     * Contains the apps and the external components they have touched.
```

src/server/AppManager.ts
graywolf336: Right, I understand that. But I think we should encapsulate the logic for the external components inside of the class `AppExternalComponentManager`. So, instead of that class having the method with the signature of `registerExternalComponents(appId: string, externalComponents: Map<string, IExternalComponent>): void` why not have it be just `registerExternalComponents(appId: string): void` and then pull from the variable `appTouchedExternalComponents` in that class?

src/server/managers/AppExternalComponentManager.ts
```diff
@@ -1,79 +1,129 @@
...
  */
 export class AppExternalComponentManager {
     /**
-     * The map that maintain the list of all registered components.
+     * The map that maintains all registered components.
+     * The key of the top map is app id and the key of inner map is the
+     * external component name.
      */
-    private providedComponents: Map<string, IExternalComponent>;
+    private providedExternalComponents: Map<string, Map<string, IExternalComponent>>;
```
graywolf336: Let's call this variable `registeredExternalComponents` and then the one below `registeredExternalComponents`. The reason the slash command uses the terminology `touched` is because we allow Apps to override system commands. But external components are new to the Apps and they can't overwrite other Apps ones or system ones.

src/server/AppManager.ts
lolimay: The better practice 👍

src/server/managers/AppExternalComponentManager.ts
lolimay: As you mentioned in another [comment](https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/156#discussion_r332281537), **overwritten** if there's an external component with the same name instead of throwing an Error. This seems to be a better solution. I will change to it.

src/server/managers/AppExternalComponentManager.ts
lolimay: Resolved in https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/150/commits/bfc6fd3fedbcb85107a83e9b0e760727710a64fe

src/server/managers/AppExternalComponentManager.ts
lolimay: Resolved in https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/150/commits/c250945c0052215f62835d6c07400c59a556265e and https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/150/commits/81d334f86757c6c76b42bebe9028f628061a7fc0

lolimay:
Hi @graywolf336 ! I also did the corresponding changes in the Rocket.Chat side. Can you please also take a look at https://github.com/RocketChat/Rocket.Chat/pull/15495? Thanks!

src/server/managers/AppExternalComponentManager.ts
```diff
@@ -1,79 +1,119 @@
...
+     * overwritten as the names provided **must** be unique.
      *
      * @param appId the id of the app
+     * @param externalComponent the external component need to be added
      */
-    public unregisterComponent(appId: string): void {
-        if (this.providedComponents.has(appId)) {
-            this.providedComponents.delete(appId);
+    public addExternalComponent(appId: string, externalComponent: IExternalComponent): void {
+        if (externalComponent.appId !== appId) {
```
graywolf336: @lolimay what would be the case this happens? Would it be that an app forgot to include it? Or they provided the wrong one? Maybe we should just set it for them instead and make things simpler?

src/server/managers/AppExternalComponentManager.ts
```diff
@@ -1,79 +1,119 @@
...
      * @param appId the id of the app
      */
-    public getExternalComponent(appId: string): IExternalComponent {
-        if (this.appTouchedComponents.has(appId)) {
-            return this.appTouchedComponents.get(appId);
+    public registerExternalComponents(appId: string): void {
+        if (!this.appTouchedExternalComponents.has(appId)) {
+            return;
         }
+        const externalComponents = this.appTouchedExternalComponents.get(appId);
```
graywolf336: What happens if this map is empty? Maybe return out early if the touched external components map is empty for some reason.

src/server/managers/AppExternalComponentManager.ts
lolimay: But if we don't add this detection, how can we ensure the develooer pass another app's appId to deceive apps-engine?

src/server/managers/AppExternalComponentManager.ts
lolimay: Thanks! Will do it as you suggested.

src/server/managers/AppExternalComponentManager.ts
graywolf336: This exact method is provided the caller's app id by the engine system (https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/145/files#diff-db7e7559fd4be39c41521a0e1b838863R10). So, that is why we can trust the `appId` provided and assign it to the `externalComponent.appId` for the App. That would ensure the external component always has the app id of whichever app provided it.

lolimay: In the current implementation, the app's developer can specify the appId of the external component freely (https://github.com/lolimay/frvr-games-rocketchat-wrapper/blob/99dae2e5fcd428ce41c945d311cc55f3653de53d/basketball-beta/BasketballBetaApp.ts#L17-L26).
Maybe we need also change it to avoid app developers' specifying the appId?

graywolf336: No, I wouldn't go that far since it would require separate models for having the `appId` limited. I would just go ahead and overwrite it inside of this method.

lolimay: Got it. Will change it as suggested 👍

lolimay: Resolved in https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/150/commits/26c638d51f1838e5ec6a971ad9409cc9df17e829.

src/server/managers/AppExternalComponentManager.ts
lolimay: Resolved in https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/150/commits/7303737cbec3e4033116a3074fddecb2d530d435.

PR #156, opened 2019-10-04
Add documentation to register method

lolimay:
resolved review: https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/145#discussion_r330056313

### DIFF
--- a/src/definition/accessors/IExternalComponentsExtend.ts
+++ b/src/definition/accessors/IExternalComponentsExtend.ts
@@ -5,5 +5,12 @@ import { IExternalComponent } from '../externalComponent';
  * component. Provided during the initialization of your App.
  */
 export interface IExternalComponentExtend {
+    /**
+     * Register an external component to the system. It actually calls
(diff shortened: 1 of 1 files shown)


src/definition/accessors/IExternalComponentsExtend.ts
```diff
@@ -5,5 +5,12 @@ import { IExternalComponent } from '../externalComponent';
  * component. Provided during the initialization of your App.
  */
 export interface IExternalComponentExtend {
+    /**
+     * Register an external component to the system. It actually calls
```
graywolf336: Let's do something like this for the method:

```
     * Register an external component to the system.
     * If you call this method twice and the component
     * has the same name as before, the first one will be
     * overwrote as the names provided **must** be unique.
```

And then this for the class:
```
/**
 * This accessor provides a method for registering external
 * components. This is provided during the initialization of your App.
 */
```

lolimay: Thanks!

PR #157, opened 2019-10-08
Update CI workflows to run on tag push

d-gubert:


rodrigok:
Is the ideia to run only on tag pushes?

graywolf336:
If we do, I think that's a bad idea...

d-gubert:
nope, but as we didn't add the tags filter before, the workflow doesn't trigger when there's a tag push.

the workflow will still run for all branches

PR #164, opened 2019-10-14
Add thread support to slash commands

d-gubert:
Apps will need to update their slash commands for them to send messages correctly aware of the thread.

Implementation provided by https://github.com/RocketChat/Rocket.Chat/pull/15574

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "typedoc": "^0.11.1"
   },
   "dependencies": {
+    "@types/node": "^12.7.12",
     "adm-zip": "^0.4.9",
     "lodash.clonedeep": "^4.5.0",
     "semver": "^5.5.0",
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
   },
   "homepage": "https://github.com/RocketChat/Rocket.Chat.Apps-engine#readme",
   "devDependencies": {
+    "@types/node": "^12.7.12",
     "@types/adm-zip": "^0.4.31",
     "@types/lodash.clonedeep": "^4.5.3",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -83,6 +83,7 @@
     "typedoc": "^0.11.1"
   },
   "dependencies": {
+    "@types/node": "^12.7.12",
```
graywolf336: 👀

d-gubert: oh shoot! `--save-dev` missing 🤦‍♂

d-gubert:
Yes, now they include the `threadId`.

I've opted to _not_ send the actual thread message object as this would add another trip to the db, and then if some app needs the data from the thread message they can call  `MessageReader.getById()`

graywolf336:
Wonder if we should add documentation to the property about that, that way people don't have to searching through issues and pull requests to get that information.

src/definition/accessors/IMessageBuilder.ts
```diff
@@ -19,6 +19,19 @@ export interface IMessageBuilder {
...
+    /**
+     * Sets the thread to which this message belongs, if any.
+     *
+     * @param threadId The id of the thread
+     */
+    setThreadId(threadId: string): IMessageBuilder;
+
+    /**
+     * Retrieves the thread to which this message belongs,
+     * if any.
```
graywolf336:
```suggestion
     * if any. Use the message reader to get the thread's message.
```

d-gubert: Updated it (see below). WDYT? 😛

PR #165, opened 2019-10-18
[ALPHA] Move host logic into apps-engine repo

lolimay:
depends on: https://github.com/RocketChat/Rocket.Chat/pull/15618


src/client/definition/EClientEmbeddedSDKActions.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * The actions provided by the AppClientEmbeddedSDK.
+ */
+export enum EClientEmbeddedSDKActions {
```
graywolf336: What is the reason for the `E` in front of the name of this enum?

lolimay: Just to let others know that it is an **E**numeration type.

graywolf336: We currently aren't doing that for other enums, so my vote would be to leave that off (https://rocketchat.github.io/Rocket.Chat.Apps-engine/enums/roomtype.html).

lolimay: Thank you, I see. Will change it.

lolimay: Resolved in https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/165/commits/f76385ca193c5b20653e189497e63357965f3e8e

src/client/AppClientUIHost.ts
```diff
@@ -0,0 +1,69 @@
...
+     */
+    public abstract async getClientUserInfo(): Promise<IClientUserInfo>;
+    /**
+     * Get the opened room's information.
+     */
+    public abstract async getClientRoomInfo(): Promise<IClientRoomInfo>;
+    /**
+     * Handle the action sent from the external component.
+     * @param action the name of the action
+     * @param id the unique id of the  API call
```
graywolf336: What is the `id`? I'm not sure I understand this part of it..

lolimay: https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/50a4c9560842e4df2928785cdb10d30a5dbfe61c/src/client/AppClientEmbeddedSDK.ts#L56-L64

In the AppClientEmbeddedSDK, we call API by specifying an action, which contains a unique key called `id`. After receiving callbacks from the host, we will delete the corresponding action by specifying the `id` from the callbacks map to indicate this action has been resolved. as shown below:
https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/50a4c9560842e4df2928785cdb10d30a5dbfe61c/src/client/AppClientEmbeddedSDK.ts#L44-L51

PR #166, opened 2019-10-19
Show detailed logs when meeting compiler errors

lolimay:
**Before:**
<kbd>before</kbd>
**After:**
<kbd>after</kbd>

### DIFF
--- a/src/server/AppManager.ts
+++ b/src/server/AppManager.ts
@@ -162,7 +162,9 @@ export class AppManager {
                 aff.setCompilerErrors(result.compilerErrors);
 
                 if (result.compilerErrors.length > 0) {
-                    throw new Error(`Failed to compile due to ${ result.compilerErrors.length } errors.`);
+                    const errors = result.compilerErrors.map(({ message }) => message).join('\n');
+
+                    throw new Error(`Failed to compile due to ${ result.compilerErrors.length } errors:\n${ errors }`);
                 }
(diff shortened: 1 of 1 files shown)


src/server/AppManager.ts
```diff
@@ -170,7 +170,9 @@ export class AppManager {
                 aff.setCompilerErrors(result.compilerErrors);
 
                 if (result.compilerErrors.length > 0) {
-                    throw new Error(`Failed to compile due to ${ result.compilerErrors.length } errors.`);
+                    const errors = result.compilerErrors.map((error) => JSON.stringify(error, null, 4)).join('\n');
```
d-gubert: Stringifying the whole error object seems to be very verbose.

Maybe just map all the `error.message`s?

d-gubert:
Theoretically, the behavior of the method hasn't changed: it's still throwing whenever there are compiler errors. So I think we wouldn't benefit that much from adding tests for the `AppManager.load` method _in this PR_ - as it would definitely not be as trivial

lolimay:
image

src/server/AppManager.ts
lolimay: resolved in https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/166/commits/46a1e0db3e09d7d7852862823c1740098778ff0c

PR #186, opened 2019-11-19
[IMPROVE] Print the apps' logs on the terminal when in development mode

lolimay:


src/server/logging/AppConsole.ts
```diff
@@ -2,14 +2,23 @@ import * as stackTrace from 'stack-trace';
...
 import { ILogEntry, ILogger, LogMessageSeverity } from '../../definition/accessors';
 import { AppMethod } from '../../definition/metadata';
+import { AppManager } from '../AppManager';
 import { ILoggerStorageEntry } from './ILoggerStorageEntry';
 
 export class AppConsole implements ILogger {
     public static toStorageEntry(appId: string, logger: AppConsole): ILoggerStorageEntry {
+        const entries = logger.getEntries();
+
+        if (logger.isDevelopmentModeEnabled) {
```
d-gubert: @lolimay I'm wondering if it wouldn't be better for us to add this part in the logging methods instead of the conversion methods. Maybe there is a scenario where the `toStorageEntry` method isn't reached? What do you think?

lolimay: I didn't put this part into the `addEntry` method because we can't get the app's information in that method 😔:
https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/8537e5ec27eb0f623ad53b0329ee07987ec9091d/src/server/logging/AppConsole.ts#L74-L95

This may cause chaos if we just print all the logs to the terminal without the app's info. Any idea that we can keep both the apps' info stayed and also all the scenarios been covered?

d-gubert: Ah! Ok, that's a very good reason! :D

d-gubert:
Ideally, the `isDevelopmentModeEnabled` should be an `async` function, since the Rocket.Chat server needs to query the database to determine if it is "in development mode". It may seem that it is not needed because _currently_ Meteor "solves" the async problem with its Fibers - but we should not rely on Meteor's features so tightly (if we ever decide to remove it, we should be prepared).

The problem is that the `isDevelopmentModeEnable` method is called during `toStorageEntry` which right now is a sync method, and changing so would incur a breaking change.

One of the possibilities would be to allow the `console.log` to be executed asynchronously. without waiting for it; it's not that important anyway. i.e.:
```
public static toStorageEntry(appId: string, logger: AppConsole): ILoggerStorageEntry {
    const entries = logger.getEntries();

    logger.isDevelopmentModeEnabled().then(shouldLog => shouldLog && entries.forEach(({ severity, args }) => console.log(`[${ severity.toUpperCase() }] ${ appId } `, args.join(' '))));

    return { ... };
}
```

But it does look rather strange.

What do you think @lolimay?

Issue #188, opened 2019-11-20
Allow to listen the 'livechat.closeRoom' callback


renatobecker:
We need to able to listen to the `livechat.closeRoom` callback through the `Apps-Engine`.
We have been working on creating a new App to provide an integration between Rocket.Chat and a customer REST API solution, so we need to be notified when a livechat room is closed.

PR #189, opened 2019-11-20
Add ILivechatRoomClosedHandler

lolimay:
# What? :boat:

Closes https://github.com/RocketChat/Rocket.Chat.Apps-engine/issues/188

# Why? :thinking:

Ref: https://github.com/RocketChat/Rocket.Chat.Apps-engine/issues/188#issue-525827629

# Links :earth_americas:

# PS :eyes:

It also depends on changes in the Rocket.Chat side: https://github.com/RocketChat/Rocket.Chat/pull/15837

### DIFF
--- a/src/server/managers/AppListenerManager.ts
+++ b/src/server/managers/AppListenerManager.ts
@@ -53,7 +53,7 @@ export class AppListenerManager {
     }
 
     // tslint:disable-next-line
-    public async executeListener(int: AppInterface, data: IMessage | IRoom | IUser): Promise<void | boolean | IMessage | IRoom | IUser> {
+    public async executeListener(int: AppInterface, data: IMessage | IRoom | IUser | ILivechatRoom): Promise<void | boolean | IMessage | IRoom | IUser | ILivechatRoom> {
         switch (int) {
             // Messages
             case AppInterface.IPreMessageSentPrevent:
--- a/src/server/managers/AppListenerManager.ts
+++ b/src/server/managers/AppListenerManager.ts
@@ -592,7 +592,7 @@ export class AppListenerManager {
             const app = this.manager.getOneById(appId);
 
             if (!app.hasMethod(AppMethod.EXECUTE_LIVECHAT_ROOM_CLOSED_HANDLER)) {
-                return;
+                continue;
             }
 
             return app.call(AppMethod.EXECUTE_LIVECHAT_ROOM_CLOSED_HANDLER,
--- a/src/server/managers/AppListenerManager.ts
+++ b/src/server/managers/AppListenerManager.ts
@@ -595,7 +595,7 @@ export class AppListenerManager {
                 continue;
             }
 
-            return app.call(AppMethod.EXECUTE_LIVECHAT_ROOM_CLOSED_HANDLER,
+            await app.call(AppMethod.EXECUTE_LIVECHAT_ROOM_CLOSED_HANDLER,
                 cfLivechatRoom,
                 this.am.getReader(appId),
                 this.am.getHttp(appId),
--- a/src/server/managers/AppListenerManager.ts
+++ b/src/server/managers/AppListenerManager.ts
@@ -95,7 +95,7 @@ export class AppListenerManager {
                 this.executePostRoomDeleted(data as IRoom);
                 return;
             // Livechat
-            case AppInterface.ILivechatRoomClosedHander:
+            case AppInterface.ILivechatRoomClosedHandler:
                 this.executeLivechatRoomClosed(data as ILivechatRoom);
                 return;
             default:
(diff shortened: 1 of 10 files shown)


renatobecker:
Guys!
I just found a problem related to the `IVisitor` interface.
As you can see [here](https://github.com/RocketChat/Rocket.Chat/blob/develop/app/apps/server/bridges/livechat.js#L89), the fields `updatedAt` and `visitorEmails` are not mandatory, but the interface is requiring them as shown below:

```
export interface IVisitor {
    id?: string;
    token: string;
    username: string;
    updatedAt: Date;
    name: string;
    department?: string;
    phone?: Array<IVisitorPhone>;
    visitorEmails: Array<IVisitorEmail>;
    customFields?: {
        [key: string]: any;
    };
}
```

What do you guys think about it? Let's fix it?
Thanks.

renatobecker:
@d-gubert and @lolimay, I just found a case even more important than the other I mentioned previously.
As you can see [here](https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/master/src/server/accessors/LivechatCreator.ts#L11) the method that allows creating a new `LivechatRoom` has 2 parameters:
- `Visitor`
- `Agent`

The problem is that the `Agent` can't be a required parameter, it must be optional because when I create a new room I may not know the agent who will attend the chat, it depends on the Livechat routing method, you know?
So, on my App I'm not able to create a new room, it's a blocker and, unfortunately, I can't move forward.
We really need to implement the `Agent` as an optional parameter ASAP, so it's a big problem right now.

src/server/managers/AppListenerManager.ts
```diff
@@ -52,7 +53,7 @@ export class AppListenerManager {
     }
 
     // tslint:disable-next-line
-    public async executeListener(int: AppInterface, data: IMessage | IRoom | IUser): Promise<void | boolean | IMessage | IRoom | IUser> {
+    public async executeListener(int: AppInterface, data: IMessage | IRoom | IUser | ILivechatRoom): Promise<void | boolean | IMessage | IRoom | IUser | ILivechatRoom> {
```
d-gubert: This is getting kinda hard to manage... In the _rich messages_ we have added one type here as well, and this tends to keep growing. We should think of a way of refactoring this in the future

src/server/managers/AppListenerManager.ts
```diff
@@ -578,4 +583,24 @@ export class AppListenerManager {
...
+        const cfLivechatRoom = Utilities.deepCloneAndFreeze(data);
+
+        for (const appId of this.listeners.get(AppInterface.ILivechatRoomClosedHandler)) {
+            const app = this.manager.getOneById(appId);
+
+            if (!app.hasMethod(AppMethod.EXECUTE_LIVECHAT_ROOM_CLOSED_HANDLER)) {
+                return;
+            }
+
+            return app.call(AppMethod.EXECUTE_LIVECHAT_ROOM_CLOSED_HANDLER,
```
d-gubert: This `return` statement is going to prevent multiple apps to handle this events - is this the expected behavior?

lolimay: Sorry, it's my fault. It should be `continue` here 😅

lolimay: early return 🙅
early continue 👈

src/server/managers/AppListenerManager.ts
```diff
@@ -578,4 +583,24 @@ export class AppListenerManager {
...
+        const cfLivechatRoom = Utilities.deepCloneAndFreeze(data);
+
+        for (const appId of this.listeners.get(AppInterface.ILivechatRoomClosedHandler)) {
+            const app = this.manager.getOneById(appId);
+
+            if (!app.hasMethod(AppMethod.EXECUTE_LIVECHAT_ROOM_CLOSED_HANDLER)) {
+                continue;
+            }
+
+            return app.call(AppMethod.EXECUTE_LIVECHAT_ROOM_CLOSED_HANDLER,
```
d-gubert:
```suggestion
            await app.call(AppMethod.EXECUTE_LIVECHAT_ROOM_CLOSED_HANDLER,
```

This `return` statement also prevents the execution of more than one app (I admit I hadn't noticed the one above 🙈). Would you agree with the proposed change?

renatobecker:
@d-gubert / @lolimay, just a quick question:

The code below is returning `undefined` when creating a new message, but it's supposed to return the `id` of the new message, right?

```
    ...
    const msg = modify.getCreator().startLivechatMessage()
    .setToken(token)
    .setVisitor(visitor)
    .setRoom(room);

    if (text) {
        msg.setText(text);
    }

    if (attachment) {
        msg.addAttachment(attachment);
    }

    return modify.getCreator().finish(msg);
```

Can you guys identify anything wrong with my code?

src/server/managers/AppListenerManager.ts
lolimay: 🥵

d-gubert:
@renatobecker nothing appears wrong with your code. It's weird that the `id` is not being returned since [the implementation](https://github.com/RocketChat/Rocket.Chat/blob/develop/app/apps/server/bridges/livechat.js#L14) seems to be taking that into account 🤔

PR #191, opened 2019-11-21
Reintroduce Bundle Installation

d-gubert:
# What? :boat:
Reintroduces Bundle Installation support for the engine

# Why? :thinking:
Due to impairment with the Rocket.Chat release cycle, the Bundle Installation logic (that included a breaking change for the app installation API)  had to be removed from the package. This PR reintroduces that behavior

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/definition/package.json
+++ b/src/definition/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@rocket.chat/apps-ts-definition",
-  "version": "1.6.0",
+  "version": "1.8.0",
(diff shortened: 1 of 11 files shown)


geekgonecrazy:
I just realized this isn't even merged?

d-gubert:
@geekgonecrazy we had to remove it since that problem we had with encrypting the app file. Work on bundles has stopped since then

PR #194, opened 2019-11-28
Update Livechat methods

d-gubert:
# What? :boat:
Deprecate inconsistent livechat methods and add new ones for data fetching

# Why? :thinking:
Some methods exposed an API requiring the App developer to know how to build a mongo query to fetch visitors and departments.

Those were deprecated and new methods to fetch such data have been added

# Links :earth_americas:

# PS :eyes:


renatobecker:
Is it possible to add a method to fetch the `Visitor` data by `Token` field? like:
- `getLivechatVisitorByToken`
It's a very useful method, we have a lot of calls to `LivechatVisitors.getVisitorByToken` on community, so for sure it would happen on the `Apps` approach very soon..

Thanks.

PR #195, opened 2019-11-29
Auto create a user for app

lolimay:
- [x] Review @Cool-fire's PR [Interface to create Bot user through Apps.engine (#133)](https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/133)
- [x] Create new role `App`. Base the permission of this role of the `bot` role, since they will be very similar
- [x] During App installation, create a user with the App
- [x] Add an "appId" field on the user that links it to the app
- [x] Prevent "App users" from logging-in on Rocket.Chat
- [x] Make the App's user the default sender for the messages sent by app

👉 The PR in the Rocket.Chat side: https://github.com/RocketChat/Rocket.Chat/pull/15896


src/server/accessors/ModifyCreator.ts
```diff
@@ -45,14 +54,24 @@ export class ModifyCreator implements IModifyCreator {
...
-    public finish(builder: IMessageBuilder | ILivechatMessageBuilder | IRoomBuilder): Promise<string> {
+    public startUser(data?: IUserCreation): IUserBuilder {
+        if (data) {
+            delete data.id;
+        }
+
+        return new UserBuilder(data);
+    }
+
+    public finish(builder: IMessageBuilder | ILivechatMessageBuilder | IRoomBuilder | IUserBuilder): Promise<string> {
```
d-gubert: This keeps haunting us... I wonder if it isn't time we stopped following this start/finish pattern 🤔

lolimay: Should we also remove the `IUserBuilder` and `UserBuilder` from the code base?

d-gubert: No, I don't think so. I think for now we just enable Apps to call a `modify.getCreator().finishUser(builder: IUserBuilder)` - this is going to make it possible to stop adding types on the `finish` method

src/definition/users/IUserCreation.ts
```diff
@@ -0,0 +1,14 @@
+export interface IUserCreation {
```
d-gubert: @lolimay this looks like a duplicate of `IUser` 🤔 Is there a technical issue preventing us from using `IUser` during user creation methods?

lolimay: Upendra and @graywolf336 has discussed it before: https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/133#discussion_r332496728 What's your idea on it?

src/server/accessors/ModifyCreator.ts
lolimay: Do you mean make different `finish*()` methods standalone?

graywolf336: @d-gubert One thing to note is that if you go that route, you will have to introduce a breaking change and all of the previous Apps will no longer work on new installations. The reason I say you will "have to" is because if you have two places for them to do it then it's not a clear and concise. If you don't want this method to be so long, then create a specific type above for it like:

`type BuilderFinisher = IMessageBuilder | ILivechatMessageBuilder | IRoomBuilder | IUserBuilder;`

Then in the finish method do `public finish(builder: BuilderFinisher): Promise<string>`

package.json
```diff
@@ -87,10 +87,12 @@
     "adm-zip": "^0.4.9",
     "cryptiles": "^4.1.3",
     "lodash.clonedeep": "^4.5.0",
+    "pre-commit": "^1.2.2",
```
graywolf336: Accidental new dependency?

graywolf336:
@d-gubert @lolimay Question. Rocket.Chat Servers are billed for their seats on Cloud and when it comes to App Subscriptions. So, us enabling Apps to create users raises the question of whether the users created by Apps are billable or not. What are your thoughts on this? Because if a developer decided to go rough, they could create a lot of users and then cause the Server to be billed for a ton of users.

src/server/AppManager.ts
```diff
@@ -393,18 +393,32 @@ export class AppManager {
...
             marketplaceInfo,
-        });
+        };
+
+        // Now that is has all been compiled, let's get the
+        // the App instance from the source.
+        const app = this.getCompiler().toSandBox(this, compiled);
+
+        // Create an app user for this app before installing it.
+        const appUserId = await this.accessorManager.getModifier(app.getID()).getCreator().getUserCreator().createAppUser(app);
```
graywolf336: The way this call is happening makes me think the App can make this call again. If that is true, is this a public method for the App and why is that so? I would think that this call would be an internal one. 🤔

lolimay: Yes, we should make this method private and also make sure it is called only once. I will change it ;) Thanks!

lolimay:
> @d-gubert @lolimay Question. Rocket.Chat Servers are billed for their seats on Cloud and when it comes to App Subscriptions. So, us enabling Apps to create users raises the question of whether the users created by Apps are billable or not. What are your thoughts on this? Because if a developer decided to go rough, they could create a lot of users and then cause the Server to be billed for a ton of users.

image

Can we use the `type` field to distinguish app users from normal users? For cloud users, the admins cannot manipulate the database directly, so this is an approachable way. For self-hosted customers, they can modify all normal users types to `app` to avoid the bills, which is a potential problem.

What do you think of it @graywolf336 @d-gubert ?

package.json
lolimay: fixed it! Thanks!

src/definition/users/IUserCreation.ts
d-gubert: I think we can replace this with `Partial<IUser>`; this makes all the fields on the `IUser` interface, which solves the problem discussed in that comment.

What do you think?

src/server/AppManager.ts
lolimay: Moved `createAppUser` to AppManager as a private method

src/definition/users/IUserCreation.ts
lolimay: Changed

src/server/accessors/ModifyCreator.ts
d-gubert: @graywolf336 I think that could be a solution for now :)

But I don't think that introducing new ways of achieving things in the engine should require breaking changes - we can always deprecate APIs and keep them around for compatibility reasons until the time comes for the next major version. Even if it means having "two places for them to do it then it's not clear and concise". This could be solved by improving our docs effort (which we definitely have to)

> Do you mean make different `finish*()` methods standalone?

@lolimay that's what I was thinking, but let's maybe leave this change for another PR

d-gubert:
@lolimay one of the pending tasks is preventing apps users from logging into Rocket.Chat, so turning a user manually into an app user via database is going to cause a lot of hassle for them. Another property of app users is that they won't have emails. So I think we could use those properties (`type = 'App', email = undefined`) to remove them from the active user count.

src/definition/users/IUser.ts
```diff
@@ -16,4 +16,9 @@ export interface IUser {
     createdAt: Date;
     updatedAt: Date;
     lastLoginAt: Date;
+    active?: boolean;
+    joinDefaultChannels?: boolean;
+    verified?: boolean;
+    requirePasswordChange?: boolean;
+    sendWelcomeEmail?: boolean;
```
graywolf336: I question whether this is something that needs to be a on a regular User. Will me updating a user after they've been created with these values result in them being set? For example, if I set `sendWelcomeEmail` on a user that has already been created will it send them an email? If not, then these properties which are only valid on the creation of a user should be on a specific interface for a creation user. So, something like `IUserCreation extends IUser` should work just find and only contain the properties. My two cents 🤷‍♂

d-gubert: Indeed, the `sendWelcomeEmail` and `joinDefaultChannels` fields definitely do not belong on the `User` schema; they're actually parameters used by the user creation method, which makes them more suited to live in the [`IUserCreationOptions`](https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/195/files#diff-8f3d6315de545e85084bdfe362dff717R3) interface

lolimay: Thanks! Already changed them as you suggested :)

src/server/bridges/IUserBridge.ts
```diff
@@ -1,9 +1,26 @@
...
+    /**
+     * Creates a user.
+     * @param data the essential data for creating a user
+     * @param appId the id of the app calling this
+     * @param options options for passing extra data
+     */
+    create(data: Partial<IUser>, appId: string, options?: IUserCreationOptions): Promise<string>;
+
+    /**
+     * Remove the app user from the system.
```
d-gubert:
```suggestion
     * Remove a user.
```

src/server/bridges/IUserBridge.ts
```diff
@@ -1,9 +1,26 @@
...
+     * @param data the essential data for creating a user
+     * @param appId the id of the app calling this
+     * @param options options for passing extra data
+     */
+    create(data: Partial<IUser>, appId: string, options?: IUserCreationOptions): Promise<string>;
+
+    /**
+     * Remove a user.
+     *
+     * @param appId the id of the app calling this
```
d-gubert:
```suggestion
     * @param user the user object to be removed
     * @param appId the id of the app executing the call
```

src/server/AppManager.ts
```diff
@@ -765,4 +783,32 @@ export class AppManager {
...
+            roles: ['app'],
+            appId: app.getID(),
+            type: UserType.APP,
+            status: 'online',
+            isEnabled: true,
+        };
+
+        return this.bridges.getUserBridge().create(userData, app.getID(), {
+            avatarUrl: app.getInfo().iconFileContent || app.getInfo().iconFile,
+            joinDefaultChannels: true,
```
rodrigok: I don't think app users should join the default channels.

d-gubert: As discussed offline, we decided to keep the `joinDefaultChannels` as `true` as it is a way for users to find out that a new app has been installed

PR #205, opened 2020-01-13
Beta (1.12.0)

d-gubert:
# What? :boat:

Changes:

- Auto create user for app during installation
- UIKit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,5 +52,6 @@ jspm_packages
 /client
 /definition
 /server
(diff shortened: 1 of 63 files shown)


d-gubert:
> @d-gubert We named an app user with the name `app.${ app.getInfo().nameSlug }` in other places, here we need to add `app.` as prefix. What do you think of it?

Not sure I follow - what do you think we need to update?

PR #212, opened 2020-02-03
[IMPROVE] Allow to check livechat online status per department

### LINKED ISSUE #211: Allow to check the status of the Omnichannel service per Department
We have a customer facing and issue in an Omnichannel App, it's related to the [isOnline](https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/339ac1815eb43fe59b0360a641c9eabc5030c5ea/src/server/accessors/LivechatRead.ts#L9) method.

Currently, it is not possible to check the status of the Omnichannel service per department, the method doesn't support this information. Therefore, it returns the status of the service regardless of the department, which provides a bad user experience.
We need to get it fixed ASAP, there is a [PR](https://github.com/RocketChat/Rocket.Chat/pull/16425) that changes the `Livechat.online` method on the Rocket.Chat repo, so we need to change it on the bridge methods too.

lolimay:
# What? :boat:

See issue #211 for the detail
# Why? :thinking:

Closes #211 
# Links :earth_americas:

https://github.com/RocketChat/Rocket.Chat/pull/16425
# PS :eyes:
n/a

### DIFF
--- a/src/definition/accessors/ILivechatRead.ts
+++ b/src/definition/accessors/ILivechatRead.ts
@@ -5,9 +5,9 @@ import { IVisitor } from '../livechat/IVisitor';
 export interface ILivechatRead {
     /**
      * Gets online status of the livechat.
-     * @param department (optional) the id of the livechat department
+     * @param departmentId (optional) the id of the livechat department
      */
-    isOnline(department?: string): boolean;
+    isOnline(departmentId?: string): boolean;
     getLivechatRooms(visitor: IVisitor, departmentId?: string): Promise<Array<ILivechatRoom>>;
     /**
      * @deprecated This method does not adhere to the conversion practices applied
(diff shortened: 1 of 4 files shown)


src/definition/accessors/ILivechatRead.ts
```diff
@@ -3,7 +3,11 @@ import { ILivechatRoom } from '../livechat/ILivechatRoom';
 import { IVisitor } from '../livechat/IVisitor';
 
 export interface ILivechatRead {
-    isOnline(): boolean;
+    /**
+     * Gets online status of the livechat.
+     * @param department (optional) the id of the livechat department
+     */
+    isOnline(department?: string): boolean;
```
renatobecker: @lolimay Maybe `departmentId` instead of `department` would be better.
What do you think?

rodrigok: I agree, `departmentId` as used below would be the correct name.

lolimay: Done, thanks!

PR #215, opened 2020-02-05
Introduce the `isOnlineAsync` method and deprecate `isOnline` on `LivechatRead`

d-gubert:
# What? :boat:

Introduce the new method `isOnlineAsync` method on the `LivechatRead` accessor and deprecate `isOnline`
# Why? :thinking:

The `isOnline` method was lacking an asynchronous signature, which was causing execution problems. In order to avoid a breaking change on the API, we've decided to introduce a new method and deprecate the current one.
# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/server/accessors/LivechatRead.ts
+++ b/src/server/accessors/LivechatRead.ts
@@ -6,8 +6,18 @@ import { ILivechatBridge } from '../bridges/ILivechatBridge';
 export class LivechatRead implements ILivechatRead {
     constructor(private readonly livechatBridge: ILivechatBridge, private readonly appId: string) { }
 
+    /**
+     * @deprecated prefer the `isOnlineAsync` method.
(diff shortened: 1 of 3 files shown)


src/server/accessors/LivechatRead.ts
```diff
@@ -6,10 +6,20 @@ import { ILivechatBridge } from '../bridges/ILivechatBridge';
 export class LivechatRead implements ILivechatRead {
     constructor(private readonly livechatBridge: ILivechatBridge, private readonly appId: string) { }
 
+    /**
+     * @deprecated prefer the `isOnlineAsync` method.
```
d-gubert:
```suggestion
     * @deprecated please use the `isOnlineAsync` method instead.
```

PR #219, opened 2020-02-06
Add container definition

sampaiodiego:
# What? :boat:
A container represents the source of an UI Kit interaction.

# Why? :thinking:
Some actions need to know from where it was triggered.

# Links :earth_americas:

# PS :eyes:


sampaiodiego:
oops.. looks like I used a branch already in use 🤔

PR #229, opened 2020-02-21
Discussion create

marceloschmidt:
# What? :boat:

- Added discussionRoom and type to messages, and added parentRoom to room;

# Why? :thinking:
This will allow us to create discussions via apps.

### DIFF
--- a/src/definition/accessors/IMessageBuilder.ts
+++ b/src/definition/accessors/IMessageBuilder.ts
@@ -1,4 +1,4 @@
-import { IMessage, IMessageAttachment } from '../messages';
+import { IMessage, IMessageAttachment, MessageType } from '../messages';
 import { RocketChatAssociationModel } from '../metadata';
 import { IRoom } from '../rooms';
 import { BlockBuilder, IBlock } from '../uikit';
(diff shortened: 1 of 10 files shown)


d-gubert:
The current implementation allows a discussion to be created _without_ a parent/origin message in the parent room. Is this something we'd like to allow? @marceloschmidt @ggazzo @sampaiodiego

marceloschmidt:
Now that you mentioned, it doesn't seem to make much sense.

marceloschmidt:
Though, do we have this type of logic here?

d-gubert:
@marceloschmidt The API as it is allows that to happen as much as the implementation on the Rocket.Chat side, so if we want to change something, we're gonna have to change in both places 😛 

I think the best thing we could do here for the developer is to provide a `createDiscussion(discussion: IRoom, origin: IMessage)` method, which would create the discussion AND the origin/parent message if needed. 

The only problem is... where do we put it? In the `IModify` interface? 🤔

sampaiodiego:
on Rocket.Chat side we have a [createDiscussion](https://github.com/RocketChat/Rocket.Chat/blob/develop/app/discussion/server/methods/createDiscussion.js#L124) method that handles everything, you can provide either a `roomId` or a `messageId`..

> I think the best thing we could do here for the developer is to provide a createDiscussion(discussion: IRoom, origin: IMessage) method, which would create the discussion AND the origin/parent message if needed.

an API like this makes more sense to me :+1:

d-gubert:
Oooh, I've recently come to realize that a message is _always_ created along with a discussion, even if the user clicks on the `Start a discussion` message action 😬 So if we were to create a bridge method that calls the `createDiscussion` method on the Rocket.Chat side, it looks like a perfect solution then?

I might be able to give this a try today

marceloschmidt:
@d-gubert did you manage to do it?

marceloschmidt:
I'm closing this one in favor of a different branch with the proposed approach.

PR #230, opened 2020-02-25
Enable Apps to create rooms without assigning a creator manually

d-gubert:
# What? :boat:

Currently, when an app wants to create a room, it is required that it provides a user to be assigned as the room creator.
This PR removes such requirement, assigning the app's default user if no other user has been provided for room creation.

# Why? :thinking:

Removing barriers to getting things done usually improves developer experience :)

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/server/accessors/ModifyCreator.ts
+++ b/src/server/accessors/ModifyCreator.ts
@@ -96,7 +96,7 @@ export class ModifyCreator implements IModifyCreator {
         return this.bridges.getLivechatBridge().createMessage(result, this.appId);
     }
 
-    private _finishRoom(builder: IRoomBuilder): Promise<string> {
+    private async _finishRoom(builder: IRoomBuilder): Promise<string> {
         const result = builder.getRoom();
         delete result.id;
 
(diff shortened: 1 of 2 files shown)


d-gubert:
@sampaiodiego do you see any problems in making this easier? Any negative impact on the product?

sampaiodiego:
need to test what happens to the rooms created by the app user when the app is uninstalled.. we have a policy to not allow a room without an owner, so _I think_ you'll not be able to uninstall the app.

sampaiodiego:
besides that, I think it is a good change 👍

d-gubert:
> need to test what happens to the rooms created by the app user when the app is uninstalled.. we have a policy to not allow a room without an owner, so I think you'll not be able to uninstall the app.

Well, that is a major blocker for this 😅 Doesn't look like there is an easy way out - We could alert the admin upon uninstallation that "It's not possible to uninstall this app because the app user is still the owner of some rooms. We've disabled the app for now, please check the rooms and manually remove them or change the owner before trying to uninstall again".

What do you think?

sampaiodiego:
maybe we should wait for https://github.com/RocketChat/Rocket.Chat/pull/16088 to decide what to do, as it will change the current behavior.

PR #239, opened 2020-03-17
[NEW] File upload validation proposal

ggazzo:
# What? :boat:

Added `IPreFileUploadAllow` that allows the rockectchat apps to allow the file validation....

the apps can validate extension, headers, find for virus or any kind of content analysis 

# Why? :thinking:
Mime types validation or content validation is a hard thing to do only with one library one single test...

# Links :earth_americas:

here an example how to validate a keynote file: 
```javascript
import { Buffer } from 'buffer';
import {
    IAppAccessors,
    ILogger,
import { IPreFileUploadAllow } from '@rocket.chat/apps-engine/definition/messages';

export class ImageFilesApp extends App implements IPreFileUploadAllow {
    constructor(info: IAppInfo, logger: ILogger, accessors: IAppAccessors) {
        super(info, logger, accessors);
    }

    public async executePreFileUploadAllow({ file, stream: buffer }: any) {
        const KEYNOTE_HEADER = Buffer.from([0x50, 0x4b, 0x03, 0x04, 0x14]);

        if (/\.(key|numbers|pages)$/.test(file.name) && buffer.compare(KEYNOTE_HEADER, 0, 5)){
            return false;
        }

        return true;
    }
}
```

# PS :eyes:

❤️ 

# PS 2

Still missing:

- [ ]  few comments adjust, and I would love introduce a way to modify the file
- [ ] create data definition (buffer and stream)

### DIFF
--- a/src/definition/messages/IPreFileUploadAllow.ts
+++ b/src/definition/messages/IPreFileUploadAllow.ts
@@ -1,0 +1,28 @@
+import { IHttp, IPersistence, IRead } from '../accessors';
+import { IMessage } from './IMessage';
+
+/**  Handler which is called to determine whether a user is allowed to send a message or not. */
+export interface IPreFileUploadAllow {
+    /**
+     * Enables the handler to signal to the Apps framework whether
+     * this handler should actually be executed for the message
+     * about to be sent.
+     *
+     * @param message The message which is being sent
+     * @param read An accessor to the environment
+     * @param http An accessor to the outside world
+     * @returns whether to run the prevent or not
+     */
+    checkPreFileUploadAllow?(message: IMessage, read: IRead, http: IHttp): Promise<boolean>;
+
+    /**
+     * Method which is to be used to prevent a message from being sent.
+     *
+     * @param message The message about to be sent
+     * @param read An accessor to the environment
+     * @param http An accessor to the outside world
+     * @param persistence An accessor to the App's persistence storage
+     * @returns whether to prevent the message from being sent
+     */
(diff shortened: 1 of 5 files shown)


engelgabriel:
@ggazzo I think this is a great start, but for multiple filtering Apps to work in conjunction, we need to have a default behavior and I believe it makes sense for the system to accept the files unless at least one App raises an objection and rejects the file. If you do the other way around, you'd need all the Apps to approve the file, but what happens if one of the Apps fails to load? 

I think the most future prof approach is that apps should return:

- ALLOW (run the next app)
- BLOCK (stop and deny)

If an App that has asked permission to filter file upload fails to initialize, the system should consider as a BLOCK, until the admin fixes the app or disables it manually. 

For clarity, if no App has blocked or registered to filter the file upload, the system would accept the file.

mitch3ls:
@engelgabriel any progress on that issue so far? We ran into a scenario where this would be very convenient

graywolf336:
Closing this in favor of the [`IPreFileUpload` interface](https://rocketchat.github.io/Rocket.Chat.Apps-engine/interfaces/uploads_iprefileupload.iprefileupload.html) and it's [`executePreFileUpload` method](https://rocketchat.github.io/Rocket.Chat.Apps-engine/interfaces/uploads_iprefileupload.iprefileupload.html#executeprefileupload).

JoshMcCullough:
@graywolf336 is there a PR that uses the [IPreFileUpload interface](https://rocketchat.github.io/Rocket.Chat.Apps-engine/interfaces/uploads_iprefileupload.iprefileupload.html) instead, or perhaps this has already been completed?

PR #242, opened 2020-03-21
new User join Event

ggazzo:
# What? :boat:

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:


d-gubert:
This has been solved in a slightly different way in https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/266

Issue #204, opened 2020-01-13
App deployment to server


smcardle:
Hi.

I created an App Engine app called "First Message" on my MBP (OSX) to deploy to our RC Docker instance running on Ubuntu 16.04 

RC version 1.2.1 - From Docker Hub
Apps Engine version 1.4.2

Then run locally on my MBP
`rc-apps deploy`

Results in the following errors:

```
Steves-MacBook-Pro:first-message stevenmcardle$ rc-apps deploy -f --url=https://app.<domain>.<tld> --username=<admin-username>
packaging your app... packaged!
And, what is the password?: *************
deploying your app... !
Error: Deployment error: Debug Failure. False expression.
    at Deploy.asyncSubmitData (/usr/local/lib/node_modules/@rocket.chat/apps-cli/lib/commands/deploy.js:100:19)
Steves-MacBook-Pro:first-message stevenmcardle$ 
```

Is this a known issue ???

If not, does anybody know why the call to `/api/apps` fails like this ???

UPDATE: *******

The log file for the Rocket.Chat docker instance reports the following:

```
app_app.1.b9vi7mn1e5s0@manager03    | Failed to resolve module: vm
app_app.1.b9vi7mn1e5s0@manager03    | Failed to resolved 1 modules for First Message v0.0.1!

```

NNesh:
I have the same problem. I am using apps-engine@1.7.0. But updating to apps-engine@1.11.0 doesn't fix it.
But it cannot find 'stream' and 'net' modules.

> packaging your app... finished!
> App packaged up at: /home/.../chat-app/dist/app_0.0.5.zip

This line appears in logs before errors:
> I20200319-13:47:00.765(5)?   /home/.../Rocket.Chat/node_modules/@types/node/child_process.d.ts [ 'events', 'stream', 'net' ]

I suppose it is due to:
> Error: /home/.../chat-app/node_modules/@rocket.chat/apps-engine/definition/accessors/IUploadRead.d.ts 
>  ›   (4,40): 
>  ›     Cannot find name 'Buffer'. Do you need to install type definitions for node? Try `npm i @types/node`.

murtaza98:
@NNesh @smcardle 
I am facing a similar error. Did u found a solution? If yes, then, please share!!
Thanks in advance.

lolimay:
Hi @murtaza98 ! Could you please share a minimal app here that can reproduce the issue. Thanks you!

[...]

PR #249, opened 2020-03-27
added support for all built-in node modules

murtaza98:
# What? :boat:

added support for all built-in node modules

# Why? :thinking:

Fixes Issue --> #204 
# Links :earth_americas:

Built-in module reference --> https://github.com/nodejs/node/tree/master/lib

# PS :eyes:


murtaza98:
@lolimay @d-gubert Please Review :)

lolimay:
Hi @murtaza98 ! Was this change by mistake https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/249/commits/b46f86c58e391e0fceb49c794adbb58ff20c02ad#diff-40cd679dee8a1f80b13930e70f0ddf50R129?

murtaza98:
@lolimay, its a mistake. Did some changes after the PR. 

Btw, I'm closing this PR based on ur suggestion [here](https://github.com/RocketChat/Rocket.Chat.Apps-engine/issues/204#issuecomment-605053528)

lolimay:
> However currently the app engine only support 4 built in modules as u can see in [this](https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/41aab31caa82b20ace7b8b354709f52d38d2bc73/src/server/misc/Utilities.ts) file. 
So what I am thinking, we can add some support for some more built in node modules which are **safe**. We won't include node modules like `vm`,  `process`, `child_process` which are a not safe from security stand point. However we can include modules like `assert`, `stream`, `timers`, etc, which are harmless.
> 
> What do u think abt this?
>  I u agree, then I will create a list of all safe modules n submit a pull request.

Hi, regarding your comment above. I will appreciated if you can help us add more *secure* node built-in modules into that file! Are you still passionate on it?

murtaza98:
@lolimay , glad u liked this idea, will start to work on it.

PR #250, opened 2020-03-27
Adds Discussion Builder

marceloschmidt:
# What? :boat:
Added Discussion Builder

# Why? :thinking:
Allows apps to create discussions

### DIFF
--- a/src/server/accessors/DiscussionBuilder.ts
+++ b/src/server/accessors/DiscussionBuilder.ts
@@ -1,155 +1,24 @@
 import { IDiscussionBuilder } from '../../definition/accessors';
 import { RocketChatAssociationModel } from '../../definition/metadata';
-import { IDiscussion, RoomType } from '../../definition/rooms';
+import { RoomType } from '../../definition/rooms';
 import { IRoom } from '../../definition/rooms/IRoom';
-import { IUser } from '../../definition/users';
+import { RoomBuilder } from './RoomBuilder';
 
-export class DiscussionBuilder implements IDiscussionBuilder {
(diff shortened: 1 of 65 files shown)


marceloschmidt:
This is related to https://github.com/RocketChat/Rocket.Chat/pull/16683/files

src/server/accessors/DiscussionBuilder.ts
```diff
@@ -0,0 +1,45 @@
...
+        this.reply = reply;
+        return this;
+    }
+
+    public getReply(): string {
+        return this.reply;
+    }
+
+    public setParentMessage(parentMessage: IMessage): IDiscussionBuilder {
+        this.parentMessage = parentMessage;
```
lolimay: + if (parentMessage && parentMessage.room) {
+    this.room.parentRoom = parentMessage.room;
+ }

Typically, the parentMessage contains the parent room info. IMO, we shouldn't provide `setParentRoom` method to developers since an unmatched pair parentRoom of parentMessage would cause some unexpecteded errors. What do you think of it? @marceloschmidt @d-gubert

marceloschmidt: This is not how Discussions work. You can have a discussion without a parentMessage.

When you set a parentMessage on a discussion, you're starting a discussion as a reply to a message: 

image

Even when you don't set a parentMessage, you still have to set a parentRoom.

lolimay: Thanks! that makes sense!

d-gubert:
Accidentally closed it - I think it was a side effect of deleting the `beta` branch 🤔

PR #146, opened 2019-09-30
Updated Typescript version to 3.6.3 #107

ashu8912:
The Current Stable Version for typescript is version 3.6.3 so updated typescript to latest version solves #107

PR #251, opened 2020-03-28
Update TypeScript to 3.8.3

ThomasVuillaume:
Set `skipLibCheck: true` to avoid the following error :  

```
Error: Deployment error: {"status":"compiler_error","messages":[{"file":"/home/vagrant/Rocket.Chat/node_modules/@types/node/index.d.ts","line":165,"character":10,"message":"/home/vagrant/Rocket.Chat/node_modules/@types/node/index.d.ts (166,11): Duplicate identifier 'IteratorResult'."},{"file":"/vagrant/Rocket.Chat.Apps-engine/node_modules/typescript/lib/lib.es2015.iterable.d.ts","line":40,"character":5,"message":"/vagrant/Rocket.Chat.Apps-engine/node_modules/typescript/lib/lib.es2015.iterable.d.ts (41,6): Duplicate identifier 'IteratorResult'."}],"success":false}
```

Tested on a Ubuntu 18 on Vagrant VM, with the following : 

```
 ➔ +-----------------------------------------------+
 ➔ |                 SERVER RUNNING                |
 ➔ +-----------------------------------------------+
 ➔ |                                               |
 ➔ |  Rocket.Chat Version: 3.0.7                   |
 ➔ |       NodeJS Version: 12.14.0 - x64           |
 ➔ |      MongoDB Version: 4.0.6                   |
 ➔ |       MongoDB Engine: wiredTiger              |
 ➔ |             Platform: linux                   |
 ➔ |         Process Port: 23563                   |
 ➔ |             Site URL: http://localhost:3000/  |
 ➔ |     ReplicaSet OpLog: Enabled                 |
 ➔ |          Commit Hash: f69d16950d              |
 ➔ |        Commit Branch: master                  |
 ➔ |                                               |
 ➔ +-----------------------------------------------+
```

To test that deployment is OK, I used the Giphy app, modified to use specific TS 3.8 ECMAScript Private Fields syntax.

Then used rc-apps to deploy it with success : 

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -238,9 +238,9 @@
       "dev": true
     },
     "@types/node": {
-      "version": "12.7.12",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-12.7.12.tgz",
-      "integrity": "sha512-KPYGmfD0/b1eXurQ59fXD1GBzhSQfz6/lKBxkaHX9dKTzjXbK68Zt7yGUxUsCS1jeTy/8aL+d9JEr+S54mpkWQ==",
+      "version": "13.9.5",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-13.9.5.tgz",
+      "integrity": "sha512-hkzMMD3xu6BrJpGVLeQ3htQQNAcOrJjX7WFmtK8zWQpz2UJf13LCFF2ALA7c9OVdvc2vQJeDdjfR35M0sBCxvw==",
       "dev": true
     },
     "@types/semver": {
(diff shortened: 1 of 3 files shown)


ThomasVuillaume:
This work was based on #107 #146

ThomasVuillaume:
Failing test is, of course, to discuss (trivial to fix if the configuration proposed is OK)

d-gubert:
Hey @ThomasVuillaume ! Thanks for the contribution! I'm going to run [Rocket.Chat's test suite](https://github.com/RocketChat/Rocket.Chat/tree/develop/tests/end-to-end/apps) with this version to test if all is good :)

graywolf336:
One thought on this that occured to me, Apps which use the new version of TypeScript will no longer be compatible with older versions of the App Engine. In my opinion, this is a breaking change. I know we all want the latest and greatest features, I want them a lot 😬, but we need to think through the strategy and consequences of this. If an App created with this version of TypeScript was attempted to be deployed on an older version of TypeScript, that version won't know what to do with the new syntax and will fail to run due to the incompatibility. 🧐🤔 

cc: @RocketChat/apps

rodrigok:
@graywolf336 makes sense and that's why we should remove the typescript compiler from inside the apps engine ASAP

graywolf336:
@rodrigok one thing to keep in mind when you guys do this, that will be a breaking change in and of itself as well. Because it won't be compatible with older versions of the Apps Engine. So, I think that would have to be in v2 or something. But I am curious and excited to see the proposal for it and what you guys have thought through and come up with 🎉

PR #255, opened 2020-04-01
App user as the default notifier

lolimay:
# What? :boat:

Make app user the default notifier when notifyUser or notifyRoom
# Why? :thinking:

Reduce dependency on rocket.cat
# Links :earth_americas:

# PS :eyes:


d-gubert:
Closed automatically when I deleted the previous beta branch. Reopening now

PR #256, opened 2020-04-03
added support for built-in node modules

murtaza98:
# What? :boat:
@lolimay following the conversation from [here](https://github.com/RocketChat/Rocket.Chat.Apps-engine/issues/204#issuecomment-605298411) (#204 #249) 

The following is the list of **secure** built-in modules

- assert,
- async_hooks, // The async_hooks module exposes functions to track asynchronous resources.
- buffer,
- crypto,
- dns,    // dns resolver
- timers,
- util,
- stream,     // for working with streaming data in Node.js
- readline,   // reading data from file
- querystring,
- string_decoder,
- url,

The following is a list of **un-secure** built-in modules which I haven't added support of

- child_process,
- cluster     // Cluster is an extensible multi-core server manager
- console     // replacement for console
- constants   // deprecated
- fs,         // file system
- os,
- process,
- worker_threads,
- v8,
- vm,
- inspector,
- module,
- perf_hooks,
- punycode,   // deprecated
- repl,
- trace_events,
- tty,
- http,
- http,
- https,
- path,
- net,
- zlib,       // compress/extract from buffer
 
For the below 4 modules I'm **not sure** whether to include them or not. Currently, I've not included them

- domain,
- events,
- globals,
- tls,        // its secure but this requires openssl installed 

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/server/misc/Utilities.ts
+++ b/src/server/misc/Utilities.ts
@@ -10,9 +10,6 @@ enum AllowedInternalModules {
     buffer,
     crypto,
     dns,
-    http,
-    https,
-    net,
     path,
     querystring,
     readline,
@@ -21,7 +18,6 @@ enum AllowedInternalModules {
     timers,
     util,
     url,
-    zlib,
 }
 
 export class Utilities {
--- a/src/server/misc/Utilities.ts
+++ b/src/server/misc/Utilities.ts
@@ -5,10 +5,23 @@ import * as vm from 'vm';
 import { ICompilerFile } from '../compiler';
 
 enum AllowedInternalModules {
-    path,
-    url,
-    crypto,
+    assert,
+    async_hooks,
     buffer,
+    crypto,
+    dns,
+    http,
+    https,
+    net,
+    path,
+    querystring,
+    readline,
+    stream,
+    string_decoder,
+    timers,
+    util,
+    url,
+    zlib,
 }
 
 export class Utilities {


lolimay:
Not sure we should provide modules such as `net`, since we actually don't want use other ways to access network excect existing APIs provided.

lolimay:
As for `zlib`, since we didn't provide filesystem r/w access, it not a good idea introduce it into the project.

murtaza98:
@lolimay thanks for the suggestion, removed support for those libraries

src/server/misc/Utilities.ts
```diff
@@ -5,10 +5,19 @@ import * as vm from 'vm';
 import { ICompilerFile } from '../compiler';
 
 enum AllowedInternalModules {
+    assert,
+    async_hooks,
+    buffer,
+    crypto,
+    dns,
```
graywolf336, line 12: This is the only one I'm unsure about, since it would allow apps to make a call outside of our app apis 🤔

src/server/misc/Utilities.ts
```diff
@@ -5,10 +5,19 @@ import * as vm from 'vm';
...
     path,
+    querystring,
+    readline,
+    stream,
+    string_decoder,
+    timers,
```
graywolf336, line 18: This is another one I don't think should be in here, since it would allow `setTimeout`, `setInterval` and those things. Which we can't guarantee the resources for an event would still be valid after the execution has happened and the engine isn't tracking the app. So, I vote to have this one removed since people might mistakenly call it and it is broken.

murtaza98:
@jschirrmacher I don't think its required now since node-modules are already supported with rocket.chat apps. More info here -> https://developer.rocket.chat/apps-engine/adding-features/npm-package-support-experimental

graywolf336:
Note about using node modules in your app and publishing them to our marketplace: you must contact us ahead of time so we can enable it on your publisher. As it is still experimental, we are gatekeeping just to ensure everything works right. Once we are out of the experimental phase then we will enable it for all publishers.

jschirrmacher:
How about using the [builtin `console` package of node.js](https://nodejs.org/docs/latest-v18.x/api/console.html)?

As of a current (1.36.0) api version of rc-apps, I get this messages:
```
packaging your app... !
[info] using TypeScript v4.9.4
2023-01-11 15:51:27.[60](https://git.tech.rz.db.de/fermata/curbside-rocket/-/jobs/108582874#L60)6 WARN  App has external module(s) as dependency
2023-01-11 15:51:27.745 WARN  App has native module(s) as dependency
 ›   Error: Failed to resolve module: console
 ```

murtaza98:
Can you show me an example of this module being used? Also do share the package.json, tslint and tsconfig file

Is it something like the native console logs that we have in node like `console.log('Some text')`? I hope its not like this because I've been using this a ton of time

jschirrmacher:
The funny thing is, there is no actual usage in my repository. An `npm ls console` doesn't show anything, and even a full text search only reveals `@jest/console` in `package-lock.json`.

PR #257, opened 2020-04-07
[NEW] Add new public method read.getAppUser

### LINKED ISSUE #243: Save app user id instead of using its username
When the app creates a user, it should store its userId in its settings and refrain from using the bot's username `app.appname`. A bot's username is editable. A userId is not.

lolimay:
# What? :boat:
Deprecate the previous method getAppUserUsername and Add recommend using new public method getAppUser

# Why? :thinking:

Closes #243 

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/definition/accessors/IUserRead.ts
+++ b/src/definition/accessors/IUserRead.ts
@@ -12,5 +12,5 @@ export interface IUserRead {
     /**
      * Gets the app user of this app.
      */
-    getAppUser(): Promise<IUser>;
+    getAppUser(appId: string): Promise<IUser>;
 }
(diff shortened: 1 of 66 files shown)


d-gubert:
This has been closed automatically when I deleted the previous beta branch. Reopening now.

src/definition/accessors/IUserRead.ts
```diff
@@ -8,4 +8,9 @@ export interface IUserRead {
     getById(id: string): Promise<IUser>;
 
     getByUsername(username: string): Promise<IUser>;
+
+    /**
+     * Gets the app user of this app.
+     */
+    getAppUser(): Promise<IUser>;
```
d-gubert, line 15: this should follow the signature from the method in the `IUserBridge` interface https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/master/src/server/bridges/IUserBridge.ts#L8

PR #266, opened 2020-05-01
New room event

d-gubert:
# What? :boat:

* Adds the new `IPreRoomUserJoined` event
* Adds the new `IPostRoomUserJoined` event
* Introduces the `AppsEngineError` class that signals prevention from apps events

# Why? :thinking:

The new event will allow apps to provide some access control to rooms.

# Links :earth_americas:

https://github.com/RocketChat/Rocket.Chat/pull/17487

# PS :eyes:

### DIFF
--- a/src/definition/AppStatus.ts
+++ b/src/definition/AppStatus.ts
@@ -51,6 +51,13 @@ export class AppStatusUtilsDef {
                 return false;
         }
     }
+
+    public isError(status: AppStatus): boolean {
+        return [
+            AppStatus.ERROR_DISABLED,
+            AppStatus.COMPILER_ERROR_DISABLED,
+        ].includes(status);
+    }
 }
 
 export const AppStatusUtils = new AppStatusUtilsDef();
--- a/src/definition/metadata/IAppInfo.ts
+++ b/src/definition/metadata/IAppInfo.ts
@@ -13,5 +13,5 @@ export interface IAppInfo {
     iconFile: string;
     /** Base64 string of the App's icon. */
     iconFileContent?: string;
-    essentials?: [AppInterface];
+    essentials?: Array<AppInterface>;
 }
--- a/src/definition/metadata/IAppInfo.ts
+++ b/src/definition/metadata/IAppInfo.ts
@@ -1,3 +1,4 @@
+import { AppInterface } from './AppInterface';
 import { IAppAuthorInfo } from './IAppAuthorInfo';
 
 export interface IAppInfo {
@@ -12,4 +13,5 @@ export interface IAppInfo {
     iconFile: string;
     /** Base64 string of the App's icon. */
     iconFileContent?: string;
+    essentials?: [AppInterface];
 }
--- a/src/definition/rooms/IPreRoomUserJoined.ts
+++ b/src/definition/rooms/IPreRoomUserJoined.ts
@@ -1,9 +1,9 @@
 import { IHttp, IPersistence, IRead } from '../accessors';
-import { AppsEngineError } from '../errors';
+import { AppsEngineException } from '../exceptions';
 import { IRoomUserJoinedContext } from './IRoomUserJoinedContext';
 
 export interface IPreRoomUserJoined {
     executePreRoomUserJoined(context: IRoomUserJoinedContext, read: IRead, http: IHttp, persistence: IPersistence): Promise<void>;
 }
 
-export class UserCannotJoinError extends AppsEngineError {}
+export class UserNotAllowedJoinException extends AppsEngineException {}
--- a/src/definition/rooms/IPreRoomUserJoined.ts
+++ b/src/definition/rooms/IPreRoomUserJoined.ts
@@ -1,16 +1,9 @@
 import { IHttp, IPersistence, IRead } from '../accessors';
 import { AppsEngineError } from '../errors';
-import { IUser } from '../users';
-import { IRoom } from './IRoom';
+import { IRoomUserJoinedContext } from './IRoomUserJoinedContext';
 
 export interface IPreRoomUserJoined {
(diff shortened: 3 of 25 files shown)


src/definition/rooms/index.ts
```diff
@@ -17,3 +17,5 @@ export {
     IPreRoomCreatePrevent,
     IPreRoomDeletePrevent,
 };
+
+export * from './IPreRoomUserJoined';
```
rodrigok, line 21: Why not follow the previous pattern?

src/server/managers/AppListenerManager.ts
```diff
@@ -618,6 +620,23 @@ export class AppListenerManager {
         }
     }
 
+    private async executePreRoomUserJoined(data: IPreRoomUserJoinedContext): Promise<void> {
+        data.room = new Room(data.room, this.manager);
```
rodrigok, line 624: The `Utilities.deepCloneAndFreeze` isn't necessary here?

src/definition/rooms/IPreRoomUserJoined.ts
```diff
@@ -0,0 +1,16 @@
...
+import { IRoom } from './IRoom';
+
+export interface IPreRoomUserJoined {
+    executePreRoomUserJoined(context: IPreRoomUserJoinedContext, read: IRead, http: IHttp, persistence: IPersistence): Promise<void>;
+}
+
+export interface IPreRoomUserJoinedContext {
+    joiningUser: IUser;
+    room: IRoom;
+    invitingUser?: IUser;
```
rodrigok, line 13: What if we name this as `actor`? This could be used later to identity who executed for other actions. Could even have more info and the user be one of the optional properties, could identify if it was a Cron action for example.

src/definition/rooms/index.ts
d-gubert: This way seems to be clearer for me than the previous one. Following an _iterative_ principle, we could apply this to all cases in the future

src/server/managers/AppListenerManager.ts
d-gubert: No - the `data` object is created by Rocket.Chat only to serve the purpose of this event. No objects inside it are referenced out of the bridges; so it wouldn't really matter if the app changes it

d-gubert: Thinking about it now, though, if an app does change the room object it might have an impact on other apps that are listening for the same event.

src/definition/rooms/IPreRoomUserJoined.ts
d-gubert: Not so sure about it... It sounds too abstract. Even though it _could_ be beneficial in some way regarding reusability (which is still unclear for me), the concepts of both actions/events you mentioned are _very_ different. If we don't have a clear understanding of what that `actor` represents and what information it shares with other contexts, it might turn out to be harmful down the road

src/definition/rooms/index.ts
rodrigok: Ok

src/definition/rooms/IPreRoomUserJoined.ts
rodrigok: I don't like the name `invitingUser` at all, it confusing to me, IMO it should be more generic indeed

d-gubert: I'm not sure why it causes confusion.

Maybe simply rewording it should be enough?

rodrigok: Please try, I'm open for suggestions and waiting. I did my already :)
I'm confused because I'm not quite sure if `invitingUser` is who invited or who was invited.

d-gubert: Changed it to `inviter`

src/definition/AppStatus.ts
```diff
@@ -51,6 +51,13 @@ export class AppStatusUtilsDef {
                 return false;
         }
     }
+
+    public isError(status: AppStatus): boolean {
+        return [
```
graywolf336, line 56: This is a micro-optimization, but wouldn't it be better to have this array created static and as a constant instead of having to be created every time this method is called?

src/definition/metadata/IAppInfo.ts
```diff
@@ -12,4 +13,5 @@ export interface IAppInfo {
     iconFile: string;
     /** Base64 string of the App's icon. */
     iconFileContent?: string;
+    essentials?: [AppInterface];
```
graywolf336, line 16: Besides being the "JavaScript" way to do it, any reason why you went with this style instead of `Array<AppInterface>`?

d-gubert: This was the first thing that crossed my mind and I haven't considered any other way of doing it until now :P It didn't feel completely right, though. Will switch, thanks

src/definition/AppStatus.ts
d-gubert: Indeed it would be slightly more efficient, indeed it is a micro-optimization. IMO it's clearer to read it that way, everything is inside the method's body.

If time comes we need this list elsewhere, no doubt we should extract to a static array. Now I'm more inclined to go with the "clearer" version

graywolf336: Interesting, I’m of the opposite opinion. I think it is more clear to read the code when it is reading a named variable/constant that expresses what it is for than to declare it in the body of the method being called.

d-gubert: Taking what you said as a rule is generally safer, but in this specific case, we have a bunch of things helping us: short method body; short method name; clear enum items named inside the array; few items inside it.

But should the situation be different, it would be worth it to go down a different path 😃

PR #267, opened 2020-05-04
[IMPROVE] Uikit allow multiple responses

lolimay:
# What? :boat:

Allow returning an array of responses while excuting `executeViewSubmitHandler`.
# Why? :thinking:

This makes closing child modal window and updating parent modal while submitting the child modal possible. Or we can only do each of them once a time.

# Links :earth_americas:

# PS :eyes:


lolimay:
Closed due to `modify.getUiController().updateModalView`.

PR #269, opened 2020-05-12
New "essentials" mechanism

d-gubert:
# What? :boat:
This is a mechanism that allows an app to register itself as "essential" to an event in the Rocket.Chat lifecycle.

If an app that is "essential" is disabled during the execution of such event, an error is thrown.

# Why? :thinking:
Currently, we have a problem regarding the state management of apps in HA setups. Each instance in the cluster has to manage the apps that are enabled/disabled, and should communicate any changes to the other instances when they happen, but this has shown to be unreliable in times when there is high usage.

This mechanism presents a way of "shielding" workspaces that require custom rules for critical events (message sending, room creating, etc.) to take place from this unpredictable problems.

# Links :earth_americas:
https://github.com/RocketChat/Rocket.Chat/pull/17656

# PS :eyes:

### DIFF
--- a/src/server/AppManager.ts
+++ b/src/server/AppManager.ts
@@ -239,21 +239,18 @@ export class AppManager {
             return;
         }
 
-        for (const rl of this.apps.values()) {
-            if (AppStatusUtils.isDisabled(rl.getStatus())) {
-                continue;
+        for (const app of this.apps.values()) {
+            if (app.getStatus() === AppStatus.INITIALIZED) {
+                this.listenerManager.unregisterListeners(app);
+                this.commandManager.unregisterCommands(app.getID());
+                this.externalComponentManager.unregisterExternalComponents(app.getID());
+                this.apiManager.unregisterApis(app.getID());
+                this.accessorManager.purifyApp(app.getID());
+            } else if (!AppStatusUtils.isDisabled(app.getStatus())) {
+                await this.disable(app.getID(), isManual ? AppStatus.MANUALLY_DISABLED : AppStatus.DISABLED);
             }
 
-            if (rl.getStatus() === AppStatus.INITIALIZED) {
-                this.listenerManager.unregisterListeners(rl);
-                this.commandManager.unregisterCommands(rl.getID());
-                this.externalComponentManager.unregisterExternalComponents(rl.getID());
-                this.apiManager.unregisterApis(rl.getID());
-                this.accessorManager.purifyApp(rl.getID());
-                continue;
-            }
-
-            await this.disable(rl.getID(), isManual ? AppStatus.MANUALLY_DISABLED : AppStatus.DISABLED);
+            this.listenerManager.releaseEssentialEvents(app);
         }
 
         // Remove all the apps from the system now that we have unloaded everything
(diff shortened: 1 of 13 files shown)


src/server/managers/AppListenerManager.ts
```diff
@@ -16,32 +17,46 @@ import {
...
 export class AppListenerManager {
     private am: AppAccessorManager;
     private listeners: Map<string, Array<string>>;
+    /**
+     * Blocked events are those who are listed in an app's
+     * "essentials" list but the app is disabled.
+     *
+     * They will throw a BlockedEventException upon call
+     */
+    private blockedEvents: Map<string, Array<string>>;
```
rodrigok, line 37:
```suggestion
    private blockedEvents: Map<string, Set<string>>;
```
Why not use set?

src/server/managers/AppListenerManager.ts
```diff
@@ -63,8 +116,18 @@ export class AppListenerManager {
         return results;
     }
 
+    public isEventBlocked(event: AppInterface): boolean {
+        const blockedEventList = this.blockedEvents.get(event);
+
+        return !!(blockedEventList && blockedEventList.length);
```
rodrigok, line 122:
```suggestion
        return !!(blockedEventList?.length);
```

src/server/AppManager.ts
```diff
@@ -796,10 +797,12 @@ export class AppManager {
             this.externalComponentManager.registerExternalComponents(app.getID());
             this.apiManager.registerApis(app.getID());
             this.listenerManager.registerListeners(app);
+            this.listenerManager.releaseEssentialEvents(app);
         } else {
             this.commandManager.unregisterCommands(app.getID());
             this.externalComponentManager.unregisterExternalComponents(app.getID());
             this.apiManager.unregisterApis(app.getID());
+            this.listenerManager.lockEssentialEvents(app);
```
rodrigok, line 805: Why use lock and block? Can we make it consistent?

src/server/managers/AppListenerManager.ts
d-gubert: It honestly did not cross my mind. Why would you consider using `Set` in this case?

src/server/managers/AppListenerManager.ts
d-gubert: Our TypeScript version does not support optional chaining :)

src/server/managers/AppListenerManager.ts
d-gubert: Given that the `Set` API provides a more convenient way to delete items, I think changing it makes sense

src/server/managers/AppListenerManager.ts
```diff
@@ -117,9 +111,9 @@ export class AppListenerManager {
     }
 
     public isEventBlocked(event: AppInterface): boolean {
-        const blockedEventList = this.blockedEvents.get(event);
+        const lockedEventList = this.lockedEvents.get(event);
 
-        return !!(blockedEventList && blockedEventList.length);
+        return !!(lockedEventList && lockedEventList.size);
```
rodrigok, line 116:
```suggestion
        return Boolean(lockedEventList?.size);
```

PR #304, opened 2020-08-26
Removing TS compiler

d-gubert:
# What? :boat:

Removing the compilation process for the apps. From this point on, we're expecting the apps package to contain their compiled versions already.

This is just part of the project - most of the work will be done on other packages.

Rocket.Chat PR: https://github.com/RocketChat/Rocket.Chat/pull/18687

# Why? :thinking:

By doing this we achieve:

* A smaller dependency tree for the package, since we will not require the `typescript` package anymore.
* A smaller memory footprint in the Rocket.Chat server, as we will not load `typescript` in runtime anymore.
* A smaller surface for compilation errors to happen when updating the Apps-Engine version in a running server.
* A shorter start-up time for the Rocket.Chat server, as it will not need to compile all installed apps anymore.
* A more flexible development experience, not tied to the `typescript` version we ship with the Apps-Engine.

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/definition/metadata/IAppInfo.ts
+++ b/src/definition/metadata/IAppInfo.ts
@@ -11,6 +11,7 @@ export interface IAppInfo {
     author: IAppAuthorInfo;
     classFile: string;
     iconFile: string;
+    implements: Array<AppInterface>;
     /** Base64 string of the App's icon. */
     iconFileContent?: string;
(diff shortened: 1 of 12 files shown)


d-gubert:
Current state: runs compiled files.

Tests have not been updated (i.e. are failing :grimacing: )

PR #305, opened 2020-08-27
[UIKit] New event handlers for the livechat widget

d-gubert:
# What? :boat:

Enables apps to listen to UIKit events that started in the livechat widget (for visitors)

# Why? :thinking:

The current UIKit events required a `user` to be always present in the event context, which is not true for events triggered in the livechat widget - the `visitor` is not considered a fully authenticated user.

The new interfaces should allow apps to listen to those events without introducing breaking changes to the API.

# Links :earth_americas:

https://github.com/RocketChat/Rocket.Chat/pull/18706
This small app has been used for testing purposes https://github.com/d-gubert/conditional-block-test

# PS :eyes:

### DIFF
--- a/src/definition/uikit/livechat/IUIKitLivechatActionHandler.ts
+++ b/src/definition/uikit/livechat/IUIKitLivechatActionHandler.ts
@@ -1,7 +1,7 @@
 import { IHttp, IModify, IPersistence, IRead } from '../../accessors';
 import { AppMethod } from '../../metadata';
 import { IUIKitResponse } from '../IUIKitInteractionType';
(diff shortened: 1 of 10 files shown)


src/definition/uikit/livechat/IUIKitLivechatActionHandler.ts
```diff
@@ -0,0 +1,23 @@
+import { IHttp, IModify, IPersistence, IRead } from '../../accessors';
+import { AppMethod } from '../../metadata';
+import { IUIKitResponse } from '../IUIKitInteractionType';
+import { UIKitLivechatBlockInteractionContext } from './UIKitLivechatInteractionContext';
+
+/** Handler for after a message is sent. */
```
d-gubert, line 6:
```suggestion
/** Handler for UIKit interactions in the livechat widget. */
```

PR #306, opened 2020-08-31
Merging "latest" release into alpha to begin next cycle

d-gubert:


d-gubert:
Changes applied via CLI

PR #332, opened 2020-10-13
User Typing indicator API

### LINKED ISSUE #303: Add support for User typing indicator
In order to provide a better UX, we need to support `user typing indicator` through the Apps-Engine.
We have been implementing integrations with `chatbot platforms` as well as other `external applications` and the ability to display that a user is typing is missing. 
For instance,  the`Salesforce Live Agent` API provides an endpoint to listen to users who are typing on `Salesforce platform`, so it would be nice to listen to that API and to upstream the event/data in the room where the external user is chatting.

That's, also, a very nice feature for omnichannel rooms that are served by chatbot agents, for instance, `Dialogflow chatbots`.

### LINKED ISSUE #330: [idea] Typing event for Rocket.Chat Bot Apps
As you are gonna move the bot logic to Rocket.Chat Apps, it is good if the bots can fire the typing event so the bot experience could be more smoothly :D

lolimay:
# What? :boat:

Added a new type indicator API:

```ts
typing(options: ITypingOptions): Promise<() => Promise<void>>;
```

# Why? :thinking:

Closes #330 
Closes #303

# Links :earth_americas:

RFC: https://forums.rocket.chat/t/rfc-30-apps-engine-add-support-for-user-typing-indicator/8684 (private)

# PS :eyes:

### DIFF
--- a/src/definition/accessors/INotifier.ts
+++ b/src/definition/accessors/INotifier.ts
@@ -3,6 +3,29 @@ import { IRoom } from '../rooms';
 import { IUser } from '../users';
 import { IMessageBuilder } from './IMessageBuilder';
 
+export enum TypingScope {
+    Room = 'room',
+}
+
+export interface ITypingOptions {
+    /**
+     * The typing scope where the typing message should be presented,
+     * TypingScope.Room by default.
+     */
+    scope?: TypingScope;
+    /**
+     * The id of the typing scope
+     *
+     * TypingScope.Room <-> room.id
+     */
+    id: string;
+    /**
+     * Timeout to stop the typing,
+     * milliseconds as unit, 15000 by default.
+     */
+    timeout?: number;
+}
+
 export interface INotifier {
     /**
      * Notifies the provided user of the provided message.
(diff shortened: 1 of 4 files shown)


lolimay:
Implementation in the RocketChat side: https://github.com/RocketChat/Rocket.Chat/pull/19228

d-gubert (changes requested):
Very good overall!

One thing I'm thinking is that we may leave the "setTimeout pattern" for the public API, but keep the bridge simply passing data around. This would mean having the bridge method receive the `options` saying whether the user is still typing, and having the `Notifier` accessor make the return cancellation-function in the `typing` method.

The reason for this is that we keep communication between the Engine and Rocket.Chat transferring simple data types (strings, objects, maybe even Buffers) instead of a complex construct like a function or an instance of a specific class, **which hold more than simple data**. Think about how you can't reliably serialize a function in order to transfer it over the network, if it was necessary.

@thassiov @sampaiodiego would like your opinions here too :)

thassiov:
I agree. Passing functions around can make it a bit more difficult to reason about the functionality, too (at least in my head it does). But it works, so 👍

lolimay:
> Very good overall!
> 
> One thing I'm thinking is that we may leave the "setTimeout pattern" for the public API, but keep the bridge simply passing data around. This would mean having the bridge method receive the `options` saying whether the user is still typing, and having the `Notifier` accessor make the return cancellation-function in the `typing` method.
> 
> The reason for this is that we keep communication between the Engine and Rocket.Chat transferring simple data types (strings, objects, maybe even Buffers) instead of a complex construct like a function or an instance of a specific class, **which hold more than simple data**. Think about how you can't reliably serialize a function in order to transfer it over the network, if it was necessary.
> 
> @thassiov @sampaiodiego would like your opinions here too :)

Thanks for the inputs :) It does make sense. I've made changes, please take another review.

d-gubert (approved):

PR #333, opened 2020-10-16
Remove doc generation from pre-push hook

d-gubert:
# What? :boat:

Refer to title 😬 
# Why? :thinking:

This is being more of a hassle than helpful as ALL doc files are updated and are not pushed at the same time with the code. We need to review the workflow around this.
# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,6 @@
   "pre-push": [
     "lint",
     "compile",
-    "unit-tests",
-    "gen-doc"
+    "unit-tests"
   ]
 }


lolimay (approved):
Will it be great to move this hook to `release a new version`?

d-gubert:
> Will it be great to move this hook to `release a new version`?

Definitely! We could have a bot open a PR with the updated docs whenever we merge to `master` for instance. We just need the time to put it there 😂

PR #342, opened 2020-10-30
Allow jobs to be scheduled along with processor registration

thassiov:
# What? :boat:

# Why? :thinking:

To be able to run tasks without the need for manually scheduling them

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/definition/accessors/ISchedulerExtend.ts
+++ b/src/definition/accessors/ISchedulerExtend.ts
@@ -6,5 +6,5 @@ export interface ISchedulerExtend {
      *
      * @param {Array<IProcessor>} processors An array of processors
      */
-    registerProcessors(processors: Array<IProcessor>): void;
+    registerProcessors(processors: Array<IProcessor>): Promise<void>;
 }
--- a/src/definition/accessors/ISchedulerExtend.ts
+++ b/src/definition/accessors/ISchedulerExtend.ts
@@ -1,5 +1,10 @@
 import { IProcessor } from '../scheduler';
 
 export interface ISchedulerExtend {
+    /**
+     * Register processors that can be scheduled to run
+     *
+     * @param {Array<IProcessor>} processors An array of processors
+     */
     registerProcessors(processors: Array<IProcessor>): void;
 }
--- a/src/definition/accessors/ISchedulerExtend.ts
+++ b/src/definition/accessors/ISchedulerExtend.ts
@@ -6,5 +6,5 @@ export interface ISchedulerExtend {
      *
      * @param {Array<IProcessor>} processors An array of processors
      */
-    registerProcessors(processors: Array<IProcessor>): Promise<void>;
+    registerProcessors(processors: Array<IProcessor>): void;
 }
--- a/src/definition/scheduler/IProcessor.ts
+++ b/src/definition/scheduler/IProcessor.ts
@@ -6,9 +6,28 @@ export interface IProcessor {
     id: string;
     /** The function that will be run on a given scheudle */
     processor: (jobContext: IJobContext, read: IRead, modify: IModify, http: IHttp, persis: IPersistence) => Promise<void>;
+    /** If provided, the processor will be configured with the setting as soon as it gets registered */
+    startupSetting?: StartupSetting;
 }
 
 /** The `data` object provided to the processor during the registering process */
 export interface IJobContext {
     [key: string]: any;
+}
+
+type StartupSetting = IOnetimeStartup | IRecurringStartup | null;
+
+interface IOnetimeStartup {
+    type: StartupType.ONETIME;
+    when: string;
+}
+
+interface IRecurringStartup {
+    type: StartupType.RECURRING;
+    cron: string;
+}
+
+export enum StartupType {
+    ONETIME = 'onetime',
+    RECURRING = 'recurring',
 }
--- a/src/definition/scheduler/IProcessor.ts
+++ b/src/definition/scheduler/IProcessor.ts
@@ -17,12 +17,12 @@ export interface IJobContext {
 
 type StartupSetting = IOnetimeStartup | IRecurringStartup | null;
 
-interface IOnetimeStartup {
+export interface IOnetimeStartup {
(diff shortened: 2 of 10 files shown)


src/definition/accessors/ISchedulerExtend.ts
```diff
@@ -1,5 +1,10 @@
 import { IProcessor } from '../scheduler';
 
 export interface ISchedulerExtend {
-    registerProcessors(processors: Array<IProcessor>): void;
+    /**
+     * Register processors that can be scheduled to run
+     *
+     * @param {Array<IProcessor>} processors An array of processors
+     */
+    registerProcessors(processors: Array<IProcessor>): Promise<void>;
```
graywolf336, line 9: Has there been a release with the return type just being `void` and not `Promise<void>`? If so, then this would be a breaking change and the original should be deprecated with a new one being a promise.

src/definition/scheduler/IRecurringSchedule.ts
```diff
@@ -1,5 +1,12 @@
+/** Represents a job that runs recurrently */
 export interface IRecurringSchedule {
+    /** The schedule's identifier */
     id: string;
+    /**
+     * When the job will be run. Must be a
+     * [human-interval](https://github.com/agenda/human-interval)
+     */
```
graywolf336, line 8: The name of this property is cron yet you're referencing a `human-interval` and I can not find any referencing in that package's readme to cron.

thassiov: I think this is a case of updating the documentation of this prop. We can attribute cron strings to it, as well as strings handled by `human-interval`, but I failed to add this in the comment

d-gubert (changes requested):

src/server/managers/AppSchedulerManager.ts
```diff
@@ -35,7 +35,11 @@ export class AppSchedulerManager {
 
             this.registeredProcessors.get(appId)[processorId] = processor;
 
-            return { id: processorId, processor: this.wrapProcessor(appId, processorId).bind(this) };
+            return {
+                id: processorId,
+                processor: this.wrapProcessor(appId, processorId).bind(this),
+                startupSetting: processor.startupSetting || null,
```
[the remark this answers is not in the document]
thassiov: Being explicit about what this property can receive, essentially. Is it preferable to leave it as a prop that holds undefined? All I want from it is a `falsy` value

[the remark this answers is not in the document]
d-gubert: For me it looks like you don't need to test anything here, then. This expression will return `null` whenever the `processor.startupSetting` is falsy, there's no need to ensure there will be a falsy value. If I understand correctly the situation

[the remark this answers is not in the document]
thassiov: I updated the code after testing and it will be `undefined`. RC server works fine with it. No need for this `or` anymore ＼⍩⃝／

src/server/bridges/IAppSchedulerBridge.ts
```diff
@@ -4,10 +4,44 @@ import {
...
+ * access to the Job Scheduler
+ */
 export interface IAppSchedulerBridge {
+    /**
+     * Register processors that can be scheduled to run
+     *
+     * @param {Array<IProcessor>} processors An array of processors
+     * @param appId the id of the app calling this
+     */
     registerProcessors(processor: Array<IProcessor>, appId: string): void;
```
d-gubert, line 18:
```suggestion
    registerProcessors(processor: Array<IProcessor>, appId: string): Promise<void>;
```

src/definition/accessors/ISchedulerExtend.ts
```diff
@@ -1,5 +1,10 @@
 import { IProcessor } from '../scheduler';
 
 export interface ISchedulerExtend {
+    /**
+     * Register processors that can be scheduled to run
+     *
+     * @param {Array<IProcessor>} processors An array of processors
+     */
     registerProcessors(processors: Array<IProcessor>): void;
```
d-gubert, line 9:
```suggestion
    registerProcessors(processors: Array<IProcessor>): Promise<void>;
```

src/definition/accessors/ISchedulerExtend.ts
d-gubert: That was not the case :)

src/definition/scheduler/IProcessor.ts
```diff
@@ -1,10 +1,42 @@
...
 export interface IJobContext {
     [key: string]: any;
 }
+
+type StartupSetting = IOnetimeStartup | IRecurringStartup;
+
+/**
+ * Processor setting for running once after being registered
+ */
+export interface IOnetimeStartup {
```
d-gubert, line 23: This interface should support a `data` property, similar to the `IOnetimeSchedule` interface, to support passing arguments over to the processor

src/definition/scheduler/IProcessor.ts
```diff
@@ -1,10 +1,42 @@
...
+    type: StartupType.ONETIME;
+    when: string;
+}
+
+/**
+ * Processor setting for running recurringly after being registered
+ */
+export interface IRecurringStartup {
```
d-gubert, line 31: This interface should support a `data` property, similar to the `IRecurringSchedule` interface, to support passing arguments over to the processor

d-gubert (changes requested):

src/definition/scheduler/IProcessor.ts
thassiov: Done!

src/definition/scheduler/IProcessor.ts
thassiov: Done!

src/definition/accessors/ISchedulerExtend.ts
graywolf336: Okay, was just making sure a release had not happened yet. :)

d-gubert (approved):

PR #358, opened 2020-12-22
Adjustments in the Permission System PR

thassiov:
# What? :boat:

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/server/AppManager.ts
+++ b/src/server/AppManager.ts
@@ -340,7 +340,7 @@ export class AppManager {
         }
         const { permissionsGranted } = app.getStorageItem();
 
-        return permissionsGranted || defaultPermissions;
+        return (permissionsGranted && permissionsGranted.length) ? permissionsGranted : defaultPermissions;
     }
 
     public async enable(id: string): Promise<boolean> {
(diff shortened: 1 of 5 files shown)


src/server/AppManager.ts
```diff
@@ -414,7 +414,7 @@ export class AppManager {
     }
 
     public async add(appPackage: Buffer, installationParameters: IAppInstallParameters): Promise<AppFabricationFulfillment> {
-        const { enable, marketplaceInfo, permissionsGranted } = installationParameters;
+        const { enable = true, marketplaceInfo, permissionsGranted } = installationParameters;
```
thassiov, line 417: Fixes the app getting disabled after installation. Having this `enable` with no default param, the app only gets initialized and not enabled.

src/server/managers/AppPermissionManager.ts
```diff
@@ -72,6 +71,7 @@ export class AppPermissionManager {
             console.error(err);
             return false;
         }
+        return true;
```
thassiov, line 74: The function does not return a positive response when if finishes, so the condition on [line 32](https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/358/files#diff-56ddd674038dcd71daa839a641d327d12e7e01984cc03bb63ac776d1601eb515L32) is always fulfilled and the actual bridged code does not get executed. Returning a positive response fixes that.

Issue #309, opened 2020-09-04
MessageBuilder should be able to add/set custom fields for message


cuonghuunguyen:
In App-engine's IMessage interface, we have the field for storing custom fields but in the MessageBuilder, there is no way to add a custom field to the Message. Could you provide a method to add custom fields to the message?

matheusbsilva137:
Hi, @meomay503 !

There is one method (addCustomField) already implemented and available in the `messageExtender` object, which can just be used after the message is sent.
The following repository may guide you through the usage of the `messageExtender`: https://github.com/RocketChat/Apps.RocketChat.Tester/tree/recipes/extending-messages

In brief, the process of sending a message returns an ID, which can be used in order to obtain the messageExtender that allows the addition of custom fields and attachments. The described process can be done as follows:

```
public async executor(context: SlashCommandContext, read: IRead, modify: IModify, http: IHttp, persis: IPersistence): Promise<void> {
    const messageId = await this.sendMessage(context, modify, 'Sending a message!');
    const messageExtender = await this.getMessageExtender(context, modify, messageId);
    const value = 1;
    const img = new ImageAttachment('https://open.rocket.chat/images/logo/logo.svg');
        
    messageExtender.addCustomField('key', value);
    messageExtender.addAttachment(img);

    await modify.getExtender().finish(messageExtender);
}
```

The complete file for the piece of code above can be found in: https://github.com/RocketChat/Apps.RocketChat.Tester/blob/recipes/extending-messages/slashcommands/ExtendSlashcommand.ts

meomay503:
Hi @matheusbsilva137, thank you for the hint. Well, your approach works and I tried it before for my use case. The problem is the latency between the moment the message is sent and the time it gets updated. 
It seems possible to create a message with custom fields and attachments by API or DDP. The current Apps Engine API also allows us to send messages with custom fields by IMessageBuilder when using with tsignore. I think the only missing thing here is the ts definition for that :D. Please correct me if you guys have any intention to disable that feature ^^

matheusbsilva137:
Hi, @meomay503!

We have performed some tests on the latency time using the `messageExtender`, and the results we obtained show that the elapsed time between sending a message and editing it depends on the quantity of new data attached to the message.
[...]

PR #363, opened 2021-01-11
[IMPROVE] Missing addCustomField method in MessageBuilder

matheusbsilva137:
# What? :boat:
 - Added the `addCustomField` method in the `MessageBuilder` class so as to allow developers to add custom fields using the `messageBuilder` object.

# Why? :thinking:
This PR is introduced in order to make the `MessageBuilder` class more complete and consistent, it has been built as a solution to the issue #309.

# Links :earth_americas:
[Issue #309 ](https://github.com/RocketChat/Rocket.Chat.Apps-engine/issues/309)

# PS :eyes:


src/server/accessors/MessageBuilder.ts
```diff
@@ -192,4 +192,17 @@ export class MessageBuilder implements IMessageBuilder {
...
+    public addCustomField(key: string, value: any): IMessageBuilder {
+        if (!this.msg.customFields) {
+            this.msg.customFields = {};
+        }
+
+        if (this.msg.customFields[key]) {
+            throw new Error(`The message already contains a custom field by the key: ${ key }`);
+        }
+
+        this.msg.customFields[key] = value;
```
graywolf336, line 205: In the documentation you say it will throw an error when the `key` contains a period but I don't see that check. I know that is something which happens from mongo but it might be better to explicitly declare it here, right?

graywolf336 (changes requested):
Potentially a missing error check. And can you cover the new lines in the tests?

src/server/accessors/MessageBuilder.ts
matheusbsilva137: Agreed! Thanks for pointing that, @graywolf336. I've also noticed that this verification is missing in some other classes in the Apps-Engine code.
I may send another pull request later so as to add these other missing exceptions.

cuonghuunguyen:
Is there any issue that blocks this PR? It's been here for a very long time

Issue #343, opened 2020-11-03
Error occurs when trying to update message using IMessageBuilder.setData() method


murtaza98:
When trying to update an existing message something like this:
```
const msgBuilder = (await this.modify.getUpdater().message(msgId, sender)).setData(newMessage);
await this.modify.getUpdater().finish(msgBuilder);
```

I'm getting an error saying 
```
Error: Invalid message, can't update a message without an id.
    at ModifyUpdater._finishMessage (/home/murtaza/github_repo/RocketChatRepos/Rocket.Chat/node_modules/@rocket.chat/apps-engine/server/accessors/ModifyUpdater.js:55:19)
    at ModifyUpdater.finish (/home/murtaza/github_repo/RocketChatRepos/Rocket.Chat/node_modules/@rocket.chat/apps-engine/server/accessors/ModifyUpdater.js:45:29)
    at PostMessageSentHandler.run (evalmachine.<anonymous>:87:48)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at D360Integration.executePostMessageSent (evalmachine.<anonymous>:22:9)
```

From what I understand, the [IMessageBuilder](https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/ab63c372a828470bdb8b6d736664300ad877d052/src/server/accessors/MessageBuilder.ts) class is deleting the `id` from the message object passed [here](https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/ab63c372a828470bdb8b6d736664300ad877d052/src/server/accessors/MessageBuilder.ts#L18)(when we call [IMessageBuilder.setData()](https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/ab63c372a828470bdb8b6d736664300ad877d052/src/server/accessors/MessageBuilder.ts#L17) method).
This behaviour is valid if we are creating an message bcs the message id is generated on server, however in case of update we don't need to delete it.

I am not sure how to fix this, any suggestions?

matheusbsilva137:
Hi, @murtaza98!

Thanks for pointing and detailedly describing this issue. In order to fix it, we have just added the **`setUpdateData` method** in the `MessageBuilder`, which allows you to do the same as the `setData` method, but it keeps the message's ID so that you can use the `ModifyUpdater` object to update the message later.
[...]

PR #364, opened 2021-01-13
[FIX] Missing a method to allow messages' updates within the MessageBuilder (#343)

matheusbsilva137:
# What? :boat:
 - Added the **`setUpdateData` method** in the `MessageBuilder`. This method is similar to the `setData` method, but it does not delete the message's ID (hence, it is possible to use the `ModifyUpdater` object to update the message later using the `messageBuilder`).

# Why? :thinking:
The `setUpdateData` method is introduced in order to allow messages' updates within the `messageBuilder` object. It has been built as a solution to the issue #343.

# Links :earth_americas:
[Issue #343](https://github.com/RocketChat/Rocket.Chat.Apps-engine/issues/343)

# PS :eyes:
The editor (user who updated the message) is also a mandatory parameter from the `setUpdateData` method, since this attribute is necessary in updates.

### DIFF
--- a/src/server/accessors/MessageBuilder.ts
+++ b/src/server/accessors/MessageBuilder.ts
@@ -21,6 +21,13 @@ export class MessageBuilder implements IMessageBuilder {
         return this;
     }
 
+    public setUpdateData(data: IMessage, editor: IUser): IMessageBuilder {
+        this.msg = data;
+        this.msg.editor = editor;
+
+        return this;
+    }
+
     public setThreadId(threadId: string): IMessageBuilder {
         this.msg.threadId = threadId;
 
--- a/src/server/accessors/MessageBuilder.ts
+++ b/src/server/accessors/MessageBuilder.ts
@@ -24,6 +24,8 @@ export class MessageBuilder implements IMessageBuilder {
     public setUpdateData(data: IMessage, editor: IUser): IMessageBuilder {
         this.msg = data;
         this.msg.editor = editor;
+        this.msg.updatedAt = new Date();
+        this.msg.editedAt = this.msg.updatedAt;
 
         return this;
     }
--- a/src/server/accessors/MessageBuilder.ts
+++ b/src/server/accessors/MessageBuilder.ts
@@ -24,8 +24,7 @@ export class MessageBuilder implements IMessageBuilder {
     public setUpdateData(data: IMessage, editor: IUser): IMessageBuilder {
         this.msg = data;
         this.msg.editor = editor;
-        this.msg.updatedAt = new Date();
-        this.msg.editedAt = this.msg.updatedAt;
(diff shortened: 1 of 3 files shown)


murtaza98:
Hey Guys,
Any update regarding when this PR will get merged?

matheusbsilva137:
> Hey Guys,
> Any estimate regarding when this PR will get merged?

We are currently in a feature freeze period of our release cycle, so we intend to merge this PR in the next few weeks (probably next week).

src/server/accessors/MessageBuilder.ts
```diff
@@ -21,6 +21,13 @@ export class MessageBuilder implements IMessageBuilder {
         return this;
     }
 
+    public setUpdateData(data: IMessage, editor: IUser): IMessageBuilder {
+        this.msg = data;
+        this.msg.editor = editor;
+
```
thassiov, line 27: Changing the message's data would require also to set the `editedAt` field from the `IMessage` object, right? So in this case adding `this.msg.editedAt = new Date();` would accomplish that. Am I on the right path?

thassiov (changes requested):
Other than the comment I left, LGTM. I'll approve as soon as it gets resolved

src/server/accessors/MessageBuilder.ts
matheusbsilva137: There's also an `msg.updatedAt` field, shouldn't the date be assigned to this field?
I'm not sure if a message update is also an edition - maybe the date should be assigned to both fields.

thassiov: I think this is the case of checking where each of them is used. Maybe this is redundant, but if it is there, must be a reason for it. I got one instance of the usage of `editedAt` in the apps engine. Can you check this? Maybe this is used in the main repo

matheusbsilva137: Both fields are used in the main repo, so I decided to update them (with the same value).

graywolf336: `editedAt` is when a message is changed by a user. `updatedAt` is whenever the system updates a message. @sampaiodiego can correct me if I'm wrong though 👀

sampaiodiego: > `editedAt` is when a message is changed by a user. `updatedAt` is whenever the system updates a message. @sampaiodiego can correct me if I'm wrong though

that's right :+1:

thassiov: @matheusbsilva137 so we should have the `editedAt` only in this case

matheusbsilva137: Great! Only the `editedAt` field is being updated now.

thassiov (approved):

PR #365, opened 2021-01-13
[FIX] Add exception to the addition of keys that contain periods

matheusbsilva137:
# What? :boat:
- Added exceptions in the `addCustomField` methods in the `MessageExtender` and `RoomExtender` classes when the key to be added contains a period.

# Why? :thinking:
The added exceptions were described in the `addCustomField` method's documentation (description) from both classes, but were not really implemented in their code.

# Links :earth_americas:

# PS :eyes:


thassiov (approved):
LGTM

PR #367, opened 2021-01-20
[FIX] Add exception to the update of required settings with empty values

matheusbsilva137:
# What? :boat:
 - Add an exception when the user attempts to update a **required** app settings' field with an empty value. Thus, required settings will not accept empty values given in the UI.

# Why? :thinking:
This PR fixes this [issue](https://app.clickup.com/t/98bbyn) on Apps Engine's side.

# Links :earth_americas:
[Issue - ClickUp](https://app.clickup.com/t/98bbyn).

# PS :eyes:

### DIFF
--- a/src/server/managers/AppSettingsManager.ts
+++ b/src/server/managers/AppSettingsManager.ts
@@ -37,6 +37,10 @@ export class AppSettingsManager {
             throw new Error('No setting found for the App by the provided id.');
         }
 
+        if (setting.value == null && rl.getStorageItem().settings[setting.id].required) {
+            throw new Error('No value given to required field.');
+        }
+
         setting.updatedAt = new Date();
         rl.getStorageItem().settings[setting.id] = setting;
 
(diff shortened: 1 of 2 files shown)


matheusbsilva137:
In our Weekly Planning Meeting, we came up with the idea of changing this PR in order to trigger the `onAppSettingsChange` event before the error is thrown.
The use case we've discussed in our meeting was related to setting a default value to a field when the user assigns an empty value to it. However, there is already one default behavior of not required fields in these scenarios -- which is to assign the `packageValue` field's value to the setting's field.
I can't think of any other use case in which that change would be valid 🤔 

Also, I feel like applying this change would bring some inconsistencies, since the `onAppSettingsChange` event would be triggered even when the setting hasn't been changed yet (in fact, the user may have just **attempted** to change the setting's value). What do you guys think? @d-gubert @lolimay @thassiov

meomay503:
may relate to #390

d-gubert:
Closing in favor of #417, which provides a less conflicting solution to the problem

PR #370, opened 2021-01-22
[IMPROVEMENT] Make the parameter on the getAppUser method optional

matheusbsilva137:
# What? :boat:
 - Updated all instances of the `getAppUser` method so as to make the `appId` parameter optional.

# Why? :thinking:
This PR implements the changes described in this [task](https://app.clickup.com/t/98bbwg), and moves closer to completely removing the `getAppUser` method's parameter in the near future.

# Links :earth_americas:
[Task - ClickUp](https://app.clickup.com/t/98bbwg)

# PS :eyes:

### DIFF
--- a/src/server/accessors/UserRead.ts
+++ b/src/server/accessors/UserRead.ts
@@ -13,11 +13,7 @@ export class UserRead implements IUserRead {
         return this.userBridge.getByUsername(username, this.appId);
     }
 
-    public getAppUser(appId?: string): Promise<IUser | undefined> {
-        if (appId === undefined) {
-            appId = this.appId;
-        }
-
+    public getAppUser(appId: string = this.appId): Promise<IUser | undefined> {
         return this.userBridge.getAppUser(appId);
     }
 }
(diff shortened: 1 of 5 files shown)


meomay503:
this would be a nice improvement (y)

src/server/accessors/UserRead.ts
```diff
@@ -13,7 +13,11 @@ export class UserRead implements IUserRead {
         return this.userBridge.getByUsername(username, this.appId);
     }
 
-    public getAppUser(appId: string): Promise<IUser | undefined> {
+    public getAppUser(appId?: string): Promise<IUser | undefined> {
```
lolimay, line 16: Use [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) here

lolimay (changes requested):

src/server/accessors/UserRead.ts
matheusbsilva137: Good suggestion! I've just implemented it.

lolimay (approved):
Nice!

PR #372, opened 2021-01-28
Permissions review

d-gubert:
# What? :boat:

After a review with some other team members, these are the suggestions that came up:

## The "general" scope and internal permissions
Having the `.general` suffix for the permissions was not a good name. The suggestion was to change it to `.manage`, but in the end, removing the suffixes altogether seemed like a better option. This way, we can have developers declare the permissions in the manifest file as `{ name: "scheduler" }`, for instance, and we maintain our internal structure by defining a "default" scope for those permissions.

We could also observe the permissions could be split into two categories: _internal_ and _system_ permissions. _Internal_ permissions were the ones that shared the `.general` suffix - they represent a feature that the app uses to manage it's own functionality internally and how it interfaces with the outside world, e.g. `scheduler`, `slashcommand`, `persistence`. These permissions do not need to be split into "read" and "write" scopes because they control "isolated" components, in a context that belongs only to the app in question. For instance, the `slashcommand` permission does not allow an app to read or write slash commands registered by other apps, only its own slash commands; also, having a `slashcommand.write` permission without having a `slashcommand.read` permission wouldn't make sense for the app, as it does not _actually_ reads and writes information residing elsewhere, but rather only register a new command and handle its execution. These internal permissions now have a single default scope, without the need for suffixing it when declaring them.

_System_ permissions are the ones that control access to resources _external_ to the app, data that does not _belong_ to it. Things like `messages`, `rooms`, server-settings, etc.

## The livechat permissions
Even though we have only one bridge for all things livechat, it manages access to more than one type of resource. Namely, we have messages, visitors, departments, etc. We split up the single "livechat" permission into multiple ones, regarding each type of resource available.

## The message.notification permission
Although internally being handled in a very different way from persistent messages, notifications are still messages in essence. Because of that, we removed the `message.notification` and require the same `message.write` permission for sending notifications (a.k.a ephemeral messages)

## The ui.interaction permission
Other _system_ permissions have "read" and "write" scopes, verbs describing the type of action the scope allows. So the `ui.interaction` permission has been changed to `ui.interact`.

### DIFF
--- a/src/server/compiler/modules/net.ts
+++ b/src/server/compiler/modules/net.ts
@@ -15,10 +15,10 @@ export const netModuleHandler = (appId: string): ProxyHandler<typeof net> => ({
             throw new ForbiddenNativeModuleAccess('net', prop);
         }
 
-        if (!AppPermissionManager.hasPermission(appId, AppPermissions.networking.general)) {
+        if (!AppPermissionManager.hasPermission(appId, AppPermissions.networking.default)) {
             throw new PermissionDeniedError({
                 appId,
-                missingPermissions: [AppPermissions.networking.general],
+                missingPermissions: [AppPermissions.networking.default],
                 methodName: `net.${prop}`,
             });
         }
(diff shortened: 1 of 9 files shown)


thassiov (approved):
LGTM

the idea of the `default` seems nice

PR #379, opened 2021-02-08
Fix missing permission checkers

d-gubert:
# What? :boat:

Some permission checkers for the Livechat bridge are missing, causing runtime errors

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:


thassiov (approved):
LGTM

PR #387, opened 2021-02-26
IPreRoomUserLeave and IPostRoomUserLeave event interfaces

lucassartor:
# What? :boat:

Added: New event interfaces for pre/post user leaving a room, `IPreRoomUserLeave` and `IPostRoomUserLeave`

# Why? :thinking:

# Links :earth_americas:

[Task](https://app.clickup.com/t/98bc63)

# PS :eyes:

### DIFF
--- a/src/definition/exceptions/UserNotAllowedException.ts
+++ b/src/definition/exceptions/UserNotAllowedException.ts
@@ -9,5 +9,7 @@ import { AppsEngineException } from '.';
  * following events:
  * - IPreRoomCreatePrevent
  * - IPreRoomUserJoined
+ * - IPreRoomUserLeave
+ * - IPostRoomUserLeave
  */
 export class UserNotAllowedException extends AppsEngineException {}
--- a/src/definition/exceptions/UserNotAllowedException.ts
+++ b/src/definition/exceptions/UserNotAllowedException.ts
@@ -10,6 +10,5 @@ import { AppsEngineException } from '.';
  * - IPreRoomCreatePrevent
  * - IPreRoomUserJoined
  * - IPreRoomUserLeave
- * - IPostRoomUserLeave
  */
 export class UserNotAllowedException extends AppsEngineException {}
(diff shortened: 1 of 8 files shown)


d-gubert:
Hey @lucassartor, can you share the app you've made to test the functionality? Thanks 🤗

d-gubert:
Also, regarding the  `IRoomUserLeaveContext` interface, is it possible to add the user who kicked the `leavingUser` when that is the case? I'm not sure that is available on the Rocket.Chat side, so maybe not possible at all

lucassartor:
> Hey @lucassartor, can you share the app you've made to test the functionality? Thanks 🤗

Sure! I'll link the compiled app that I've made to test it.
[user-leave_0.0.1.zip](https://github.com/RocketChat/Rocket.Chat.Apps-engine/files/6069439/user-leave_0.0.1.zip)

lucassartor:
> Also, regarding the  `IRoomUserLeaveContext` interface, is it possible to add the user who kicked the `leavingUser` when that is the case? I'm not sure that is available on the Rocket.Chat side, so maybe not possible at all

I've tried to create a `kicker` in the `IRoomUserLeaveContext` interface but I was unable to get that information from the Rocket.Chat side. 
I think that's currently not possible, but I'm not 100% sure, maybe I got it wrong on the implementation.

d-gubert:
@RocketChat/backend halp! can some of you confirm the information above is not available?

sampaiodiego:
we have `beforeLeaveRoom` and `afterLeaveRoom` callbacks: 
https://github.com/RocketChat/Rocket.Chat/blob/f7700514e50287179b4a93f4eb76747dce83ac47/app/lib/server/functions/removeUserFromRoom.js#L10

is that what you guys are missing?

d-gubert:
@sampaiodiego  we're wondering if it is possible to get the user who kicked another user from a room. For instance, if `user A` kicks `user B` from the room, the event will trigger with the room and `user B` in the context - is there a way to tell that `user A` was the one who made that action?

sampaiodiego:
that's little messy :see_no_evil: for event `afterRemoveFromRoom` we have that information.. but not sure if it triggers on all events needed.

sampaiodiego:
if that is not enough I'd say we should start using a new structure built on top of services to expand that. we can discuss how that should look like.

d-gubert:
Ok, we'll move on with this implementation now, then. After our discussions we can add the new field if it makes sense. Thanks @sampaiodiego !

d-gubert (approved):

PR #392, opened 2021-03-09
Enable typing of the request body and the response body

graywolf336:
# What? :boat:
- Enables app developers to type their request and responses

# Why? :thinking:
- So developers can have a better development experience
- I wanted this while working on an App 🙈

### DIFF
--- a/src/definition/accessors/IHttp.ts
+++ b/src/definition/accessors/IHttp.ts
@@ -6,15 +6,15 @@ import { IRead } from './IRead';
  * Based off of: https://github.com/meteor-typings/meteor/blob/master/1.4/main.d.ts#L869
  */
 export interface IHttp {
-    get(url: string, options?: IHttpRequest): Promise<IHttpResponse>;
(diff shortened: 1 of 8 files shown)


d-gubert:
It concerns me that adding this will give a false impression of "validity" to the response of a request. Might make people think that the Apps-Engine actually validated the payload in the `data` field against the provided type.

At the end of the day, as a developer, you will still need to validate that the `data` object has the fields you are expecting it to have, even if your typing says it's there. So I guess the "manual" solution of checking and typecasting/type-predicates might still be better, as it makes sure you don't forget to validate data. E.g.:

```typescript
type KnownData = { knownField: string };

function isKnownData(data: unknown) data is KnownData {
    return !!data?.knownField;
}

// Inside some function....

const { data } = await http.get('https://my.api.com/some/route');

if (!isKnownData(data)) throw new Error();

// Typescript is now able to tell `data` is of type KnownData

PR #399, opened 2021-03-22
[IMPROVE] Refactor Bridges

d-gubert:
Using a `Proxy` to check for permissions in the bridges ended up being _too loosely coupled_ with the bridges themselves. For instance, to create a new bridge, one is required to:

* Create a new interface in the bridges folder,
* Add that bridge to the Bridge union type in the AppBridges file (so it is recognised to be a bridge),
* Then create a new permission checker for that bridge in the permissions folder, noting that both have different names;
* and those three components will be brought together dynamically by the `AppPermissionManager`, which wraps the bridge with a `Proxy` at runtime when the engine gets it via the `getNewBridge()` method.

Another example of this is that the `Proxy` used in the `AppPermissionManager` relies on the name of the bridge to be exactly the same both in the host implementation and the permission checker, which is very loosely coupled and error-prone.
Conversely, the `AppPermissionManager` is too tied up to the concept of bridges by providing a method to wrap only bridge objects.

Basically, _there is no mechanism in the code linking these components_, even though they are conceptually linked.

## Solution Proposal
One possible way to tie up these knots is by refactoring the bridges to be classes instead of interfaces. This will allow us to require the host system to extend these classes when implementing the bridges, which will provide us with a more reliable way of asserting that the host's implementation is correct and will also help the host with this implementation. Another benefit will be the possibility of colocating the permission checking with the actual bridge definition, by employing the [Template Method pattern](https://en.wikipedia.org/wiki/Template_method_pattern).

For example, let's take the message bridge: instead of having an `IMessageBridge` interface, an `AppMessageBridge` permission checker and the `AppPermissionManager` tying them up with a `Proxy`, we could have:
```typescript
abstract class AppMessageBridge extends BridgeBase {
    /**
     * Here we defined the template methods for each one of the current
     * ones defined by the bridge. These template methods are going to do
     * the actual permission checking just before running the methods implemented
     * by the host system.
     *
     * For the sake of brevity, I'll introduce the example of just one template
     * method below.
     */
    public doCreate(message: IMessage, appId: string): Promise<string> {
        if (!AppPermissionManager.hasPermission(appId, AppPermissions.message.write)) {
            throw new PermissionDeniedError({
                appId,
                missingPermissions: [AppPermissions.message.write],
            });
        }


        return this.create(message, appId);
    }
    /**
     * These are the same methods defined in the IMessageBridge interface today
     * and these are the ones that should be implemented by the host system
     */
    abstract protected create(message: IMessage, appId: string): Promise<string>;
    abstract protected getById(messageId: string, appId: string): Promise<IMessage>;
    abstract protected update(message: IMessage, appId: string): Promise<void>;
    abstract protected notifyUser(user: IUser, message: IMessage, appId: string): Promise<void>;
    abstract protected notifyRoom(room: IRoom, message: IMessage, appId: string): Promise<void>;
    abstract protected typing(options: ITypingDescriptor, appId: string): Promise<void>;

### CODE REVIEW

**Review by d-gubert — COMMENTED** [review]
* Do we still need the bridge interfaces? For instance `IUserBridge` or `ILivechatBridge`. Considering now we have other definitions of the bridges in the form of abstract classes, it looks to me that we can safely remove their interface counterparts.

* Some bridges don't require permissions to execute their actions (`AppActivationBridge` for instance). Do they still need the `doStuff()` and `stuff()` method definitions?

* With the new changes, no error is thrown when an app tries to do something it doesn't have permission to. This can make troubleshooting much harder.

@thassiov what do you think?

File: src/server/bridges/MessageBridge.ts [hunk 1]
```diff
@@ -0,0 +1,84 @@
...
+    protected abstract create(message: IMessage, appId: string): Promise<string>;
+    protected abstract update(message: IMessage, appId: string): Promise<void>;
+    protected abstract notifyUser(user: IUser, message: IMessage, appId: string): Promise<void>;
+    protected abstract notifyRoom(room: IRoom, message: IMessage, appId: string): Promise<void>;
+    protected abstract typing(options: ITypingDescriptor, appId: string): Promise<void>;
+    protected abstract getById(messageId: string, appId: string): Promise<IMessage>;
+
+    private hasReadPermission(appId: string): boolean {
+        if (AppPermissionManager.hasPermission(appId, AppPermissions.message.read)) {
+            console.log('message read permissions has', appId);
```
felipetomm, line 60: @d-gubert , that's an necessary console.log or we can remove them in a future PR?

d-gubert: Oops. Yeah, I guess we can remove it in the next release. Thanks for the heads up :)

PR #403, opened 2021-04-12
Add `skipImmediate` property to the IRecurringSchedule interface

thassiov:
# What? :boat:

Adds a property lets the developer decide whether they want the first iteration of a recurring task to be executed immediately instead of waiting for the defined interval to complete

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/definition/scheduler/IRecurringSchedule.ts
+++ b/src/definition/scheduler/IRecurringSchedule.ts
@@ -8,6 +8,10 @@ export interface IRecurringSchedule {
      * or a number.
      */
     interval: string | number;
+    /**
+     * Whether to let the first iteration to execute as soon as the task is registered
+     */
+    skipImmediate?: boolean;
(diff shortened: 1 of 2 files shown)


lolimay (approved):

src/definition/scheduler/IRecurringSchedule.ts
```diff
@@ -8,6 +8,10 @@ export interface IRecurringSchedule {
      * or a number.
      */
     interval: string | number;
+    /**
+     * Whether to let the first iteration to execute as soon as the task is registered
```
d-gubert, line 12: Bit of rewording, wdyt?

```suggestion
     * Whether the recurring job should start immediately or wait for the interval
```

d-gubert (changes requested):

d-gubert (approved):

Issue #397, opened 2021-03-18
[Feature Request] Get info about who is installing / uninstalling an app


murtaza98:
The feature request is to add a feature via which the apps-engine should pass the `IUser` object of the user who has installed the app or has Uninstalled an app. Apps-Engine could pass this info within the current `onEnable()` and `onDisable()` method itself or maybe there could be new methods for installation / uninstallation and this info could be shared there. 

For our omnichannel apps, there are certain use-cases where we need this info - so for example - within the new Twitter app - we could send a message to the user who has installed the app with the relevant OAuth link via which they can authorise our Twitter app to access info from their twitter account.

PR #404, opened 2021-04-14
[NEW] App `onInstall` event

lucassartor:
# What? :boat:

**Added:** `onInstall` event method.

# Why? :thinking:

# Links :earth_americas:

[ClickUp Task](https://app.clickup.com/t/98bbxj)
Fixes  https://github.com/RocketChat/Rocket.Chat.Apps-engine/issues/397

# PS :eyes:
Complements https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/407

### DIFF
--- a/src/server/AppManager.ts
+++ b/src/server/AppManager.ts
@@ -414,7 +414,7 @@ export class AppManager {
         return true;
     }
 
-    public async add(appPackage: Buffer, installationParameters: IAppInstallParameters): Promise<AppFabricationFulfillment> {
+    public async add(appPackage: Buffer, installationParameters: IAppInstallParameters, user: IUser): Promise<AppFabricationFulfillment> {
         const { enable = true, marketplaceInfo, permissionsGranted } = installationParameters;
 
         const aff = new AppFabricationFulfillment();
@@ -474,7 +474,7 @@ export class AppManager {
             // If an error occurs during this, oh well.
         });
 
-        await this.installApp(created, app);
+        await this.installApp(created, app, user);
 
         // Should enable === true, then we go through the entire start up process
         // Otherwise, we only initialize it.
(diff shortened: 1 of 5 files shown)


src/server/AppManager.ts
```diff
@@ -733,6 +735,31 @@ export class AppManager {
...
+
+        try {
+            await app.call(AppMethod.ONINSTALL, read, http, persistence, user);
+
+            result = true;
+        } catch (e) {
+            const status = AppStatus.ERROR_DISABLED;
+
+            if (e.name === 'NotEnoughMethodArgumentsError') {
+                console.warn('Please report the following error:');
```
lolimay, line 752: Don't simply use `console.warn`, use `app.getLogger().warn` instead. Please append the original error to the log as well.

lucassartor: Ok I've fixed it. I used `console.warn` because other methods in this file use it, my bad!

d-gubert (changes requested):

src/server/AppManager.ts
```diff
@@ -414,7 +414,7 @@ export class AppManager {
         return true;
     }
 
-    public async add(appPackage: Buffer, installationParameters: IAppInstallParameters): Promise<AppFabricationFulfillment> {
+    public async add(appPackage: Buffer, installationParameters: IAppInstallParameters, user: IUser): Promise<AppFabricationFulfillment> {
```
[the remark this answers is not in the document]
lucassartor: It really does. I've fixed it, thanks!

src/server/AppManager.ts
```diff
@@ -733,6 +735,31 @@ export class AppManager {
         return this.enableApp(storageItem, app, true, isManual, silenceStatus);
     }
 
+    private async installApp(storageItem: IAppStorageItem, app: ProxiedApp, user: IUser): Promise<boolean> {
+        let result: boolean;
```
thassiov, line 740: Maybe lose this variable and return a boolean directly in the `try/catch` block? We won't run any code after it, after all. Does it make sense?

thassiov (commented):
LGTM. There's just the context thing left to put in the method's argument, right?
**EDIT**: nvm, for some odd reason I saw only 3 files instead of 5, and the context were in the two that disappeared :thinking:

d-gubert (approved):

src/server/AppManager.ts
d-gubert: Let's review this during the beta period

PR #406, opened 2021-04-19
[NEW] Get departments

d-gubert:
# What? :boat:

Add a new method that allows apps to fetch all departments in the server.
# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/definition/livechat/IDepartment.ts
+++ b/src/definition/livechat/IDepartment.ts
@@ -2,6 +2,13 @@ export interface IDepartment {
     id: string;
     name?: string;
     email?: string;
+    description?: string;
+    offlineMessageChannelName?: string;
+    requestTagBeforeClosingChat?: false;
+    chatClosingTags?: Array<string>;
+    abandonedRoomsCloseCustomMessage?: string;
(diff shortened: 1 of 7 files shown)


src/definition/livechat/IDepartment.ts
```diff
@@ -2,6 +2,13 @@ export interface IDepartment {
     id: string;
     name?: string;
     email?: string;
+    description?: string;
+    offlineMessageChannelName?: string;
+    requestTagBeforeClosingChat?: false;
+    chatClosingTags?: Array<string>;
+    abandonedRoomsCloseCustomMessage?: string;
+    waitingQueueMessage?: string;
+    departmentsAllowedToForward?: string;
```
d-gubert, line 11: @renatobecker please review the data types and if these fields are really optional, as I don't have much familiarity with that domain 😛

renatobecker: Everything looks good @d-gubert !
In fact, all of these fields are optional because, except description field, all the others are EE fields.

PR #407, opened 2021-04-19
[NEW] App `onUninstall` event

lucassartor:
# What? :boat:

**Added:** `onUninstall` event method.

# Why? :thinking:

# Links :earth_americas:

[ClickUp Task](https://app.clickup.com/t/98bbxj)
Fixes  https://github.com/RocketChat/Rocket.Chat.Apps-engine/issues/397

# PS :eyes:
Complements https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/404

### DIFF
--- a/src/definition/App.ts
+++ b/src/definition/App.ts
@@ -5,6 +5,7 @@ import {
     IEnvironmentRead,
     IHttp,
     ILogger,
+    IPersistence,
     IRead,
 } from './accessors';
 import { AppStatus } from './AppStatus';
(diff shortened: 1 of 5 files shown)


src/definition/App.ts
```diff
@@ -161,6 +163,19 @@ export abstract class App implements IApp {
         return;
     }
 
+    /**
+     * Method which is called when the App is uninstalled and it is called one single time.
+     * If the App is disabled this method will NOT be called.
```
graywolf336, line 168:
```suggestion
     * This method will NOT be called when an App is getting disabled. This method will only be called whenever the App is being uninstalled and removed from Rocket.Chat. This method will be called when an App is getting uninstalled and the App was disabled. 
```

graywolf336 (changes requested):
Not sure about wording or the order of operations.

d-gubert (approved):

PR #417, opened 2021-05-13
[NEW] Add ON_PRE_SETTING_UPDATE hook to apps

lolimay:
# What? :boat:

This PR introduces a new `onPreSettingUpdate` hook into apps. In this way, app developers can validate new setting update before it takes effects.

# Why? :thinking:

[RFC 37](https://forums.rocket.chat/t/rfc-37-apps-engine-add-on-pre-setting-update-hook-to-apps/11156)

# Links :earth_americas:

# PS :eyes:


lolimay:
```typescript
import {
    IAppAccessors, IConfigurationExtend, IConfigurationModify, IEnvironmentRead, IHttp, ILogger, IRead,
import { ISetting, SettingType } from '@rocket.chat/apps-engine/definition/settings';
import { Toast } from '@rocket.chat/apps-engine/definition/exceptions';

export class AppsSettings extends App {
    constructor(info: IAppInfo, logger: ILogger, accessors: IAppAccessors) {
        super(info, logger, accessors);
    }

    public async initialize(configurationExtend: IConfigurationExtend, environmentRead: IEnvironmentRead): Promise<void> {
        configurationExtend.settings.provideSetting({
            id: 'Telegram',
            public: true,
            type: SettingType.SELECT,
            packageValue: 'none',
            value: 'none',
            values: [
                {
                    key: 'none',
                    i18nLabel: 'none',
                },
                {
                    key: 'name',
                    i18nLabel: 'name',
                },
                {
                    key: 'username',
                    i18nLabel: 'username',
                },
                {
                    key: 'nickname',
                    i18nLabel: 'nickname',
                },
            ],
            i18nLabel: 'telegram_agents_display_name',
            i18nDescription: 'telegram_agents_display_name_description',
            required: true,
        });
    }

    public async onPreSettingUpdate(setting: ISetting, configurationModify: IConfigurationModify, read: IRead, http: IHttp): Promise<ISetting> {
        // Practice: We cannot get oldSetting in this way
        const oldSetting = await read.getEnvironmentReader().getSettings().getById(setting.id);
        console.log({ oldSetting: oldSetting.value, newSetting: setting.value });

        // Your custom setting change validation
        switch (setting.id) {
            case 'Telegram': {
                switch (setting.value) {
                    case 'none':
                        throw new Toast.Error('The value of the setting Telegram cannot be `none`');
                }
                break;
            }
        }
        return setting;
    }
}
```
The app for testing: 
[apps-settings_0.0.1.zip](https://github.com/RocketChat/Rocket.Chat.Apps-engine/files/6515466/apps-settings_0.0.1.zip)

src/definition/App.ts
```diff
@@ -197,6 +197,19 @@ export abstract class App implements IApp {
...
+    /**
+     * Method which is called before a setting which belongs to this App is going to be updated
+     * by an external system and not this App itself. The setting passed is the newly updated one.
+     *
+     * @param setting the setting which is going to be updated
+     * @param configurationModify the accessor to modifiy the system
+     * @param reader the reader accessor
+     * @param http an accessor to the outside world
+     */
+    public async onPreSettingUpdate(setting: ISetting, configurationModify: IConfigurationModify, read: IRead, http: IHttp): Promise<ISetting> {
```
d-gubert, line 209: I would think the setting's "old value" could be passed as a parameter here, maybe as a `SettingUpdateContext` object as a first parameter. The Apps-Engine already has this value in its scope, so providing it as a parameter of the hook will prevent apps from having to reach the DB to grab it

src/server/ProxiedApp.ts
```diff
@@ -94,6 +94,10 @@ export class ProxiedApp implements IApp {
             logger.error(e);
             logger.debug(`'${method}' was unsuccessful.`);
 
+            if (e instanceof Toast.Error) {
```
d-gubert, line 97: Ideally, this case would use the same inheritance of `AppsEngineException` as seen below, since that "pattern" is used in some other places as well. Is this use case different in a way that prevents it from using that?

d-gubert (changes requested):

src/definition/exceptions/Toast.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * The custom exceptions from apps
+ *
+ * It's used to display toasts to users that * a _known_
+ * exception has happened during the execution of the apps.
+ */
+class ErrorToast extends Error { }
+export const Toast = {
```
d-gubert, line 8: Other exceptions for similar purposes follow a different naming pattern, as `FileUploadNotAllowedException` for instance (https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/refactor-bridges/src/definition/exceptions/FileUploadNotAllowedException.ts).

d-gubert (changes requested):

src/definition/exceptions/Toast.ts
d-gubert: I still think the name is not appropriate for the error. Naming it `Toast` with an internal `ErrorToast` prop ties it too much to how the host system will implement the error display _and_ can be misleading in the sense that it might make you think other errors _won't be shown as a toast_, which is not the case.

I'd suggest following the `FileUploadNotAllowedException` pattern, and name the exported class as `InvalidSettingValueException` for instance

d-gubert (changes requested):

d-gubert (approved):

Issue #425, opened 2021-06-09
[BUG] IServerSettingRead getValueById fn not working as expected for BOOLEAN/STRING Setting


Shailesh351:
```
public async getValueById(id: string): Promise<any> {
        const set = await this.settingBridge.getOneById(id, this.appId);

        if (typeof set === 'undefined') {
            throw new Error(`No Server Setting found, or it is unaccessible, by the id of "${id}".`);
        }

        return set.value || set.packageValue;
    }
```

This is how `getValueById` function gets value.

The issue is with the BOOLEAN settings. When `packageValue` (default value) is set to true and `value` (current value) if false,
It returns true because of `OR ||`.

Similar issue when setting value is `''` for STRING Setting.

### Expected behaviour 
It should return false in the above case.

### Possible Solution
```
public async getValueById(id: string): Promise < any > {
	const set = await this.settingBridge.getOneById(id, this.appId);

	if (typeof set === 'undefined') {
		throw new Error(`No Server Setting found, or it is unaccessible, by the id of "${id}".`);
	}

	if (set.value === undefined || set.value === null) {
		return set.packageValue;
	}

	return set.value;
}
```

PR #426, opened 2021-06-09
[FIX] Update getValueById for properly getting value for BOOLEAN and STRING Setting

Shailesh351:
# Fixes https://github.com/RocketChat/Rocket.Chat.Apps-engine/issues/425

# What? :boat:

- Updated ISettingRead getValueById
- Updated IServerSettingRead getValueById

# Why?

#### Earlier Behaviour
```
public async getValueById(id: string): Promise<any> {
        const set = await this.settingBridge.getOneById(id, this.appId);

        if (typeof set === 'undefined') {
            throw new Error(`No Server Setting found, or it is unaccessible, by the id of "${id}".`);
        }

        return set.value || set.packageValue;
    }
```

The issue is with the BOOLEAN settings. When `packageValue` (default value) is set to true and `value` (current value) if false,
It returns true because of `OR ||`.

Similar issue when setting value is `''` for STRING Setting.

#### Behaviour after this PR

Properly returns BOOLEAN and STRING setting

### DIFF
--- a/src/server/accessors/ServerSettingRead.ts
+++ b/src/server/accessors/ServerSettingRead.ts
@@ -17,7 +17,11 @@ export class ServerSettingRead implements IServerSettingRead {
             throw new Error(`No Server Setting found, or it is unaccessible, by the id of "${id}".`);
         }
 
-        return set.value || set.packageValue;
+        if (set.value === undefined || set.value === null) {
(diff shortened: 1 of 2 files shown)


d-gubert (approved):
Thank you for your contribution! :D

PR #431, opened 2021-07-01
Fix wrong permission on department read method

d-gubert:
# What? :boat:

Fix wrong permission check in the `doFindDepartmentByIdOrName` method

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:


cuonghuunguyen:
seems multiple permission is not added

PR #435, opened 2021-08-19
[FIX] Make username property optional for IVisitor

murtaza98:
# What? :boat:

The `IVisitor` interface of the apps engine currently makes the usernames field mandatory, however this isn't required as on the backend we have a logic in-place to generate a new username if no usernames are provided
# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/docs/interfaces/livechat_ivisitor.ivisitor.html
+++ b/docs/interfaces/livechat_ivisitor.ivisitor.html
@@ -198,7 +198,7 @@
 				</section>
 				<section class="tsd-panel tsd-member tsd-kind-property tsd-parent-kind-interface">
 					<a name="username" class="tsd-anchor"></a>
-					<h3>username</h3>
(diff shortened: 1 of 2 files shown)


murtaza98:
Hi @d-gubert Could you please review this PR ?

d-gubert:
Hey @murtaza98 ! This PR is actually introducing a breaking change 😬  Any app that reads the `username` property of a visitor **will not compile anymore** after updating to this change. So we cannot merge it as-is.

I'm assuming the intent here is to allow apps to not assign a username to the visitor when they are being created, right? For this case, you could actually change the method that creates a visitor so it accepts a `Partial<IVisitor>` parameter instead of a `IVisitor` parameter - this change would be backward compatible

ggazzo:
I would suggest having subtype of IVisitor without username, only for specific cases. if not you will be forced to test in every single place of the app if there is or not username

d-gubert:
Closing it for now as it is breaking

PR #438, opened 2021-08-25
[NEW] Add a bridge to the INet interface

kirillk77:
# What? :boat:
This pull request contains an implementation of the bridge to the INet interface. 

# Why? :thinking:
The INet interface adds the ability to create Rocket.Chat applications that uses binary network protocols. At the moment the Apps-Engine limits applications using the one and only HTTP protocol.


d-gubert:
It is already possible to use the native `net` module in the Apps-Engine, you can require it like so `import net = require('net');` :)

PR #439, opened 2021-08-25
[NEW] Add a bridge to the ITimers interface

cybem:
# What? :boat:
This pull request contains an implementation of the bridge to the ITimers interface.

# Why? :thinking:
The ITimers interface is very useful to implement custom protocols, circuit breakers, throttlers, complex actions and so on.

### DIFF
--- a/src/definition/accessors/IAppAccessors.ts
+++ b/src/definition/accessors/IAppAccessors.ts
@@ -1,4 +1,4 @@
-import { IEnvironmentRead, IHttp, IRead } from '../../definition/accessors';
+import { IEnvironmentRead, IHttp, IRead, ITimers } from '../../definition/accessors';
 import { IApiEndpointMetadata } from '../api';
 
 export interface IAppAccessors {
@@ -6,4 +6,5 @@ export interface IAppAccessors {
     readonly reader: IRead;
     readonly http: IHttp;
     readonly providedApiEndpoints: Array<IApiEndpointMetadata>;
+    readonly timers: ITimers;
 }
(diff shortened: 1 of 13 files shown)


cybem:
> Hey @cybem! Thank you for taking the time to make this PR :)
> 
> One thing that's not clear to me based on the implementation is how you intend to use this new bridge in the app. Can you provide an example?

Hello, Douglas @d-gubert!

We use it in that manner:
```    
   private delay(ms: number) {
        const setTimeout = this.timers.call('setTimeout');
        return new Promise((resolve) => setTimeout(resolve, ms));
    }
```
But we can adapt source code of our app to more elegant solution of the bridge.

Thank you for your feedback!

d-gubert:
> But we can adapt source code of our app to more elegant solution of the bridge.

We're considering allowing access to the functions directly via global scope, much like you'd have in a pure NodeJS environment. I think it would solve your use case as well, right? 

Usually we create bridges for apps to access resources that are directly related to Rocket.Chat, such as data entities like messages and rooms; or features that are tightly coupled to the database, like the persistence layer or scheduler. 

These functions from the `timers` module are provided globally by NodeJS, so we could pass them directly to the app as well.

What do you think?

cybem:
> What do you think?

Hello Douglas @d-gubert,

Yes, sure. It would be nice.

Thank you!

ankar84:
Hello, @d-gubert 
As you discussed with @cybem `timers` module was globally available at late 2021. But now looks like that module not available again.
We see that `timers` module is imported in Apps-Engine here https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/7af3781e5170a898030328b0502e75d9c990af8d/src/server/misc/Utilities.ts#L3
But here https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/eaec68e8179255d81364c88ab68cacbd3f120c87/src/server/compiler/modules/index.ts it not used!
So, looks like a regression critical for us.
Can you check that?
Thank you!

PR #441, opened 2021-09-03
[NEW] Enable require of some native modules and global functions

thassiov:
# What? :boat:

Allows the app to include code that requires the following modules:
- `http` (requires `networking` permission)
- `https` (requires `networking` permission) 
- `zlib`
- `util`
- `os` (as an empty object, only to fulfill some npm dependencies' requirements, as exposing it would lead to security problems)

Also, injects the following language features inside the app's context so developers can make use of:
- [timers](https://nodejs.org/dist/latest-v16.x/docs/api/timers.html) (`setInterval`, `setImmediate`, etc)
- `Buffer`

# Why? :thinking:

As we are allowing external dependencies in apps, some internal functionalities and libs are needed for the apps to function properly.
# Links :earth_americas:

# PS :eyes:


thassiov:
[github-app.zip](https://github.com/RocketChat/Rocket.Chat.Apps-engine/files/7108644/github-app.zip)

tester app

src/server/ProxiedApp.ts
```diff
@@ -51,6 +53,8 @@ export class ProxiedApp implements IApp {
 
     public makeContext(data: object): vm.Context {
         return vm.createContext(Object.assign({}, {
+            ...timers,
+            Buffer,
```
d-gubert, line 57: Would be good to have a sort of `buildDefaultAppContext` that always injects these properties into the global scope of apps

d-gubert (changes requested):

src/server/ProxiedApp.ts
thassiov: Done

graywolf336 (approved):

PR #443, opened 2021-09-23
[NEW] Storage Strategy and NPM dependencies

thassiov:
# What? :boat:

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:


src/definition/accessors/ILivechatRead.ts
```diff
@@ -27,4 +28,9 @@ export interface ILivechatRead {
     getLivechatVisitorByToken(token: string): Promise<IVisitor | undefined>;
     getLivechatVisitorByPhoneNumber(phoneNumber: string): Promise<IVisitor | undefined>;
     getLivechatDepartmentByIdOrName(value: string): Promise<IDepartment | undefined>;
+    /**
+     * @experimental we do not encourage the wider usage of this method,
+     * as we're evaluating its performance and fit for the API.
+     */
+    _fetchLivechatRoomMessages(roomId: string): Promise<Array<IMessage>>;
```
graywolf336, line 35: Can this be turned into an `AsyncIterableIterator`?

thassiov: It could be done, but I don't think it would work well in the scenario it tries to address. The use case for it is getting all the messages from a livechat room in one go so the user/developer could forward it to an archive outside rocket.chat (like a crm tool). Returning a list is just simpler because the volume of messages is usually low when dealing with livechat

graywolf336: That is understandable. However, it could still potentially block the rocket.chat server if the discussions are any length. Going to an async approach would prevent potential blocking of the node event loop. 🤔

d-gubert (approved):

PR #449, opened 2021-10-29
[FIX] App update flow causes errors in HA setups

d-gubert:
# What? :boat:

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:


thassiov (approved):
LGTM!

PR #442, opened 2021-09-06
[NEW] Omnichannel source type

d-gubert:
# What? :boat:

Add the new fields in Omnichannel rooms to identify the source of the room

# Why? :thinking:

Fields have been added to Rocket.Chat types

# Links :earth_americas:

https://github.com/RocketChat/Rocket.Chat/pull/23090
# PS :eyes:

PR #459, opened 2021-12-09
[NEW] adding custom sidebar icon method

AllanPazRibeiro:
# What? :boat:

This PR is about a feature that allows the App Developer to send a custom icon (SVG) for the room's sidebar.

With that in mind, this PR adds the following stuff:
- Method to set the icon name;
- Add the room icon source type;

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:


src/server/accessors/RoomBuilder.ts
```diff
@@ -8,7 +8,7 @@ export class RoomBuilder implements IRoomBuilder {
     protected room: IRoom;
     private members: Array<string>;
 
-    constructor(data?: Partial<IRoom>) {
+     constructor(data?: Partial<IRoom>) {
```
graywolf336, line 11: Why the extra space?

graywolf336:
My concern with SVG is we open ourselves to the common SVG attacks, unless there’s a way to get around this

src/server/accessors/RoomBuilder.ts
AllanPazRibeiro: ops, fixing this

AllanPazRibeiro:
> 

Hi @graywolf336, thank you for your comments, regarding the security issues, it is dealt with on the frontend by the below actions:

- All the potential injection attributes are removed.
- And no inline script is executed.

For a more detailed explanation of how the frontend deal with it, you can reach @ggazzo

src/definition/rooms/IRoom.ts
```diff
@@ -23,4 +24,11 @@ export interface IRoom {
     customFields?: { [key: string]: any };
     parentRoom?: IRoom;
     livechatData?: { [key: string]: any };
+    source?: {
```
graywolf336, line 27: This `source` property is so generic that someone who does not know Rocket.Chat's internals would not know what this is. In fact, I personally have no idea what it is referring to. What's the purpose of this? Can it be named something better and then mapped inside of Rocket.Chat?

src/definition/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rocket.chat/apps-ts-definition",
-  "version": "1.28.0-alpha",
+  "version": "1.29.0-alpha.0",
```
graywolf336, line 3: Please revert this change. I know it is added automatically but it does not need to be committed by this PR.

src/definition/rooms/IRoom.ts
AllanPazRibeiro: I agree that this name is too generic but I just decided to use the same property name as it used on the omnichannel room on Rocket.chat project as you can see on this [PR](https://github.com/RocketChat/Rocket.Chat/pull/23912)

src/server/accessors/RoomBuilder.ts
```diff
@@ -150,4 +150,9 @@ export class RoomBuilder implements IRoomBuilder {
     public getRoom(): IRoom {
         return this.room;
     }
+
+    public setSidebarIcon(sidebarIcon: string): IRoomBuilder {
+        this.room.source.sidebarIcon = sidebarIcon;
```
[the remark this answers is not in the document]
AllanPazRibeiro: Since this property is not mandatory I decided to set it as an optional property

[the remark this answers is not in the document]
graywolf336: Right, but what I am saying is that you're setting `source.sidebarIcon` and `source` could be undefined. Thus, you need to ensure that the object exists before you set a value on it.

src/definition/rooms/IRoom.ts
graywolf336: Right, I understand that. However, what I am saying is that we should map it to a name that actually provides meaning instead of further compounding the confusion and generics.

AllanPazRibeiro:
@graywolf336 I just commit a PR about your comments

src/definition/rooms/IRoom.ts
```diff
@@ -23,4 +24,11 @@ export interface IRoom {
     customFields?: { [key: string]: any };
     parentRoom?: IRoom;
     livechatData?: { [key: string]: any };
+    sideBarIconSource?: {
```
murtaza98, line 27: Considering this is an omnichannel property, so can we add this `ILivechatRoom` instead of the base room object?

murtaza98 (changes requested):
Sorry a bit late to the review party here :sweat_smile: 

I had one piece of feedback. On all our omnichannel apps, we use [this](https://rocketchat.github.io/Rocket.Chat.Apps-engine/interfaces/accessors_ilivechatcreator.ilivechatcreator.html#createroom) method to create an omnichannel room, so it would be awesome if we could add an extra param to this method to define the source of the channel as well as when it is getting created... This way we don't have to make another call to DB to update the source post room creation

src/definition/rooms/IRoom.ts
graywolf336: From what I understand, this property should not be exposed actually.

graywolf336: > I mean, the idea of that prop is to not be exposed so others can modify (we didn't enforce this), currently, the apps-engine sets the value on room creation based on apps info (but this happens only for omnichannel rooms) since that prop should not exists on other room types

@KevLehman

AllanPazRibeiro: @ggazzo just created a PR with some modifcations on this PR https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/462#pullrequestreview-832976836

felipetomm:
> Sorry a bit late to the review party here sweat_smile
> 
> I had one piece of feedback. On all our omnichannel apps, we use [this](https://rocketchat.github.io/Rocket.Chat.Apps-engine/interfaces/accessors_ilivechatcreator.ilivechatcreator.html#createroom) method to create an omnichannel room, so it would be awesome if we could add an extra param to this method to define the source of the channel when it is getting created... This way we don't have to make another call to DB to update the source post room creation

Wow!!! That's very awesome if you can add this info. Is an very important info to get metrics about customers channels, like others behaviors. At moment we identify channels by customFields of every Omnichannel Apps (and this is so rong)... A time ago i make an question about this [here](https://github.com/RocketChat/Rocket.Chat.Apps-engine/pull/442#issuecomment-916292451) when Apps Source Type were implemented.
Thanks for note this @murtaza98 . We are looking forward to it :)

src/definition/rooms/IRoom.ts
KevLehman: I'm a bit confused with the implementation. It basically allows N number of channel icons (different for each room). That means an app would be able to set a different icon for every room it creates (which is a feature that can generate confusion to the end user as well)

From what I remember, an app would be able to use one icon and then all rooms created by the app would use that icon (if the app wanted it). Was this changed? 👀 

(Additionally, will we limit the ability to set custom room icons to only omnichannel rooms? I know this is what we need right now, but, with PR 462, the code validations would need to be removed if apps engine wanted to allow this in the future)

ggazzo: we don't know if the same app can't have more than one integration, today we develop one by one, but it's up to the developer. The app will decide this when creating the room, this is a little less "plug and play", but at the same time more flexible, it involves less metadata and the behavior is also simpler. Today we are going to store the data in the room, otherwise we would need to store the information in the app, collect and evaluate this data, I don't think we have that prepared (I'm not an apps-engine expert, but I think it would take an infinite amount of work and a code more "coupled")

Today we are not interested in changing room icons to other types, we think this is a mess actually, but who knows, maybe in the future this 'source' field can live in all rooms not just omnichannel.

> I'm a bit confused with the implementation. It basically allows N number of channel icons (different for each room). That means an app would be able to set a different icon for every room it creates (which is a feature that can generate confusion to the end user as well)
> 
> From what I remember, an app would be able to use one icon and then all rooms created by the app would use that icon (if the app wanted it). Was this changed? 👀
> 
> (Additionally, will we limit the ability to set custom room icons to only omnichannel rooms? I know this is what we need right now, but, with PR 462, the code validations would need to be removed if apps engine wanted to allow this in the future)

AllanPazRibeiro:
Folks,

I just committed a few modifications along with @ggazzo modifications to add the source as a parameter on LiveChat's create room method as @murtaza98 mentioned.

And I decided to keep the set and get methods to modify the source icon props for future use cases.

PS: I know that would be better to have only one set and one get method do source but the source, in this case, has a different type (IRoom != ILivechatRoom)

marking whose is in this thread:
@graywolf336 @ggazzo @felipetomm @murtaza98 @KevLehman @d-gubert

murtaza98 (approved):
Thanks @AllanPazRibeiro It looks good to me!! Really looking forward to the release of this one :rocket:

ggazzo (approved):

PR #461, opened 2021-12-10
[NEW] Add new event for pre email

d-gubert:
# What? :boat:

Adds definitions and listener for new event `IPreEmailSent`, that allows for manipulation of email messages sent and interruption of transmission if needed
# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/definition/email/IEmailDescriptor.ts
+++ b/src/definition/email/IEmailDescriptor.ts
@@ -1,0 +1,12 @@
+
+export interface IEmailDescriptor {
+    from?: string | undefined;
+    to?: string | Array<string> | undefined;
+    cc?: string | Array<string> | undefined;
+    bcc?: string | Array<string> | undefined;
+    replyTo?: string | Array<string> | undefined;
+    subject?: string | undefined;
+    text?: string | undefined;
+    html?: string | undefined;
+    headers?: Record<string, string> | undefined;
+}
--- a/src/definition/email/IPreEmailSent.ts
+++ b/src/definition/email/IPreEmailSent.ts
@@ -1,0 +1,25 @@
+import { IEmailDescriptor, IPreEmailSentContext } from '.';
+import { IHttp, IModify, IPersistence, IRead } from '../accessors';
+import { AppMethod } from '../metadata';
+
+/**
+ * Event interface that allows apps to
+ * register as a handler of of the `IPreEmailSent`
+ * event.
+ *
+ * This event is trigger before the mailer sends
+ * an email.
+ *
+ * To prevent the email from being sent, you can
+ * throw an error with a message specifying the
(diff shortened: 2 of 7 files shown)


src/definition/email/IEmailDescriptor.ts
```diff
@@ -0,0 +1,12 @@
+
+export interface IEmailDescriptor {
+    from?: string | undefined;
+    to?: string | Array<string> | undefined;
```
graywolf336, line 4: Is the last `undefined` needed? I haven't touched TS in a while but I always thought having the question mark automatically added that

src/definition/email/IPreEmailSent.ts
```diff
@@ -0,0 +1,25 @@
...
+ * reason for rejection.
+ */
+export interface IPreEmailSent {
+    [AppMethod.EXECUTE_PRE_EMAIL_SENT](
+        context: IPreEmailSentContext,
+        read: IRead,
+        http: IHttp,
+        persis: IPersistence,
+        modify: IModify,
+    ): Promise<IEmailDescriptor>;
```
graywolf336, line 24: I am assuming this can modify the message? If so, maybe add the suffix `Modify` like we have for the messages and rooms. That way people know it's a method that can mutate the data

src/definition/email/IEmailDescriptor.ts
d-gubert: The `?` and `undefined` do have different meanings: adding `?` makes it optional to _include_ that property when creating an object that follows an interface; while `undefined` allows for the inclusion of the property with `undefined` as a value.

So, if we removed the `? ` from the type, it would be mandatory to include `{ to: undefined }` at least in the object. On the other hand, if we removed `undefined`, _it wouldn't be possible to pass pass that property as `undefined`_, i.e. `{ to: undefined }`.

But we did mostly copy it from here https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/meteor/email.d.ts#L4-L12 :joy:

graywolf336: Makes sense to me. Been so long I've forgot about all the inner workings and specifics of TypeScript 🙈

src/definition/email/IPreEmailSent.ts
d-gubert: We're moving away from that naming convention for events, to consolidate them into `Pre` or `Post` only. So this can modify _and_ prevent the event by throwing an error.

graywolf336: Sounds like a good thing for v2 ;)

ggazzo (approved):

PR #479, opened 2022-01-14
[NEW] Adding bridge to change both status and status text

AllanPazRibeiro:
# What? :boat:

This PR adds a bridge method to change the status and status text.

# Why? :thinking:

This functionality will help the apps to set custom status better.

# Links :earth_americas:

# PS :eyes:


thassiov:
Looks good, but I was thinking that we should retrofit this module with permission checks. Thoughts?

thassiov (approved):

PR #482, opened 2022-01-20
[IMPROVE] Add label prop to source in livechat room

d-gubert:
# What? :boat:

Add label prop to source in livechat room
# Why? :thinking:

This label will be shown in the room info
# Links :earth_americas:

# PS :eyes:


felipetomm:
@d-gubert what you think about an prop that be used for represent a channel of livechat room?
I feel that structure isn't general when we have rooms started by Widgets x Apps. If i need to get the channel that an visitor is comming, we need to validate if an room is from widget or app. If comes from app, we need to verify his alias. But alias that's not represents the "channel" right? Is only an name designed by the app.
Currently we know that an app is responsible for one channel, but is not an rule.. I can create an app that support one or more channels.

PR #487, opened 2022-02-09
[NEW] Adding the method to authenticate and the method to get the token

AllanPazRibeiro:
# What? :boat:

This PR adds the auth callback method to save the token and the method to retrieve the token from the persistence layer;

# Why? :thinking:

Because the app would need the token to access the user data.

# Links :earth_americas:

# PS :eyes:


src/server/oauth2/OAuth2Client.ts
```diff
@@ -89,18 +140,71 @@ export class OAuth2Client {
...
+                        state,
+                    ),
+                    new RocketChatAssociationRecord(
+                        RocketChatAssociationModel.MISC,
+                        `${this.config.alias}-oauth-connection`,
+                    ),
+                ],
+            );
+
+            return {
```
d-gubert, line 195: I'm wondering if we should have a callback here for the App to do something when the user authorizes the connection. Something along the lines of 

```
autorizationCallback: (token: IOAuthToken, user: IUser /*, accessors */): Promise<{
    responseContent?: string; // Content to be shown to the user maybe
} | undefined>
```

This function would have to be sent as a parameter to the `setup` method and stored as a property of the `config` object.

What do you think? @AllanPazRibeiro @thassiov

d-gubert (changes requested):

src/server/oauth2/OAuth2Client.ts
```diff
@@ -34,41 +45,67 @@ export class OAuth2Client {
...
 
-        await configuration.settings.provideSetting({
-            id: `${this.config.alias}-oauth-client-id`,
-            type: SettingType.STRING,
-            public: true,
-            required: true,
-            packageValue: '',
-            i18nLabel: `${this.config.alias}-oauth-client-id`,
-        });
+        await Promise.all([
```
d-gubert, line 56: Cool trick :)

d-gubert (approved):

PR #491, opened 2022-02-16
[NEW] OAuth2 supporting methods

AllanPazRibeiro:
# What? :boat:

- This PR adds the supporting methods to revoke/refresh the token access.

# Why? :thinking:

- These methods allow the apps to manage the authorization flow

# Links :earth_americas:

# PS :eyes:


src/server/oauth2/OAuth2Client.ts
```diff
@@ -207,4 +219,118 @@ export class OAuth2Client {
...
+        userId,
+        persis,
+    }: {
+        token: string,
+        expiresAt: number,
+        refreshToken: string,
+        userId: string,
+        persis: IPersistence,
+    }) {
+        await persis.createWithAssociations(
```
d-gubert, line 305: We can replace the `await` for a `return` - as this is the last statement of the function, `await`ing it essentially does nothing, as there will be nothing to run afterwards. By using a `return`, we're deferring the `await`ing to the calling method

src/server/oauth2/OAuth2Client.ts
```diff
@@ -207,4 +219,118 @@ export class OAuth2Client {
...
+                new RocketChatAssociationRecord(
+                    RocketChatAssociationModel.MISC,
+                    `${this.config.alias}-oauth-connection`,
+                ),
+            ],
+        );
+    }
+
+    private async removeToken({ userId, persis }: { userId: string, persis: IPersistence} ) {
+        await persis.removeByAssociations([
```
d-gubert, line 325: We can replace the `await` for a `return` - as this is the last statement of the function, `await`ing it essentially does nothing, as there will be nothing to run afterwards. By using a `return`, we're deferring the `await`ing to the calling method

src/server/oauth2/OAuth2Client.ts
```diff
@@ -207,4 +219,118 @@ export class OAuth2Client {
...
+
+        const url = new URL(this.config.revokeTokenURI);
+
+        for (const tokenInfo of tokensInfo) {
+            url.searchParams.set('token', tokenInfo.token);
+
+            await this.app.getAccessors().http.post(url.href);
+        }
+
+        this.removeToken({ userId: user.id, persis });
```
d-gubert, line 288: When calling functions/methods that return promises, we should either `await` or `return` _always_ - here, for instance, even if a method calling `revokeUserAccessToken` uses an `await` **it will end up not waiting until the `removeToken` method finishes running to return the control**, because this call to `removeToken` simply creates a Promise but doesn't act on it

src/server/oauth2/OAuth2Client.ts
```diff
@@ -174,30 +185,31 @@ export class OAuth2Client {
...
+                    status: statusCode,
+                    content: '<div style="display: flex;align-items: center;justify-content: center; height: 100%;">\
+                                <h1 style="text-align: center; font-family: Helvetica Neue;">\
+                                    Oops, something went wrong, please try again or in case it still does not work, contact the administrator.\
+                                </h1>\
+                            </div>',
+                };
+            }
+
+            this.saveToken({
```
d-gubert, line 199: When calling functions/methods that return promises, we should either `await` or `return` _always_ - here, this function will return the object below before the execution of `saveToken` finishes running

d-gubert (changes requested):
Apart from the comments provided, many of the methods lack a return type, please provide those as well

d-gubert (approved):

PR #496, opened 2022-04-04
[FIX] Faulty redirect_uri path concatenation

thassiov:
# What? :boat:

During the OAuth process, the `redirect_uri` wasn't being formatted properly, making the login process fail. 

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:


graywolf336 (changes requested):
Unfortunately, the siteUrl in Rocket.Chat sometimes does have a slash in it. So, you might have to add a little bit of code to enforce that there is only one `/` when you are combining them.

tapiarafael (approved):
LGTM

graywolf336 (approved):

PR #499, opened 2022-04-12
[NEW] Add new events after user has been created, updated and deleted

tapiarafael:
# What? :boat:
Add definition and listeners for new events `IPostUserCreated`, `IPostUserUpdated` and `IPostUserDeleted`

# Why? :thinking:

# Links :earth_americas:
[Task](https://app.clickup.com/t/22wqd4p)
[Rocket.chat PR](https://github.com/RocketChat/Rocket.Chat/pull/25165)

# PS :eyes:


src/server/managers/AppListenerManager.ts
```diff
@@ -1196,4 +1203,58 @@ export class AppListenerManager {
 
         return descriptor;
     }
+
+    private async executePostUserCreated(data: IUserContext): Promise<void> {
+        const context = Object.freeze(data);
```
d-gubert, line 1208: `Object.freeze` performs a shallow freeze only, in these cases it's better to use `Utilities.deepFreeze` to ensure inner objects are also frozen

src/definition/users/IUserContext.ts
```diff
@@ -0,0 +1,6 @@
+import { IUser } from './IUser';
+
+export interface IUserContext {
+    user: IUser;
+    performedBy?: IUser;
+}
```
d-gubert, line 6: It would be good to add some comments here explaining app devs what is the purpose of the interface and the properties

src/definition/users/IUserUpdateContex.ts
```diff
@@ -0,0 +1,7 @@
+import { IUser } from './IUser';
+
+export interface IUserUpdateContext {
+    user: IUser;
+    performedBy?: IUser;
+    previousData?: IUser;
+}
```
d-gubert, line 7: It would be good to add some comments here explaining app devs what is the purpose of the interface and the properties

d-gubert (changes requested):

thassiov (approved):
LGTM

d-gubert (approved):

PR #501, opened 2022-04-27
[NEW] Adding interfaces to add action button on user dropdown menu

AllanPazRibeiro:
# What? :boat:
- Added: Enum option to user dropdown

# Why? :thinking:

This PR is necessary for the Rocket.chat will use this interface to add the button on the user dropdown menu

# Links :earth_americas:

# PS :eyes:


AllanPazRibeiro:
> Do we have an example of this? I think it would be neat to see an example App (side note: would be cool to see a test/example app for new features, that way people can see it in action)

yes, we have a doc: https://developer.rocket.chat/apps-engine/fundamentals-of-apps/uikit/action-buttons

PR #502, opened 2022-04-28
[NEW] Add message react, follow and pin message events

tapiarafael:
# What? :boat:

Added definition and listeners for new events IPostMessageReacted, IPostMessagePinned and IPostMessageFollowed
Added properties in the Message interface to give more context about pinned messages.

# Why? :thinking:

# Links :earth_americas:

[Task](https://app.clickup.com/t/22wqd51)
[Rocket.chat PR](https://github.com/RocketChat/Rocket.Chat/pull/25337)

# PS :eyes:

### DIFF
--- a/src/definition/messages/IMessage.ts
+++ b/src/definition/messages/IMessage.ts
@@ -25,4 +25,8 @@ export interface IMessage {
     parseUrls?: boolean;
     customFields?: { [key: string]: any };
     blocks?: Array<IBlock>;
+    starred?: Array<{ _id: string }>;
+    pinned?: boolean;
+    pinnedAt?: Date;
+    pinnedBy?: { _id: string, username: string };
 }
--- a/src/definition/messages/IMessage.ts
+++ b/src/definition/messages/IMessage.ts
@@ -1,6 +1,6 @@
 import { IRoom } from '../rooms';
 import { IBlock } from '../uikit';
-import { IUser } from '../users';
+import { IUser, IUserLookup } from '../users';
 import { IMessageAttachment } from './IMessageAttachment';
 import { IMessageFile } from './IMessageFile';
 import { IMessageReactions } from './IMessageReaction';
@@ -28,5 +28,5 @@ export interface IMessage {
     starred?: Array<{ _id: string }>;
     pinned?: boolean;
     pinnedAt?: Date;
-    pinnedBy?: { _id: string, username: string };
+    pinnedBy?: IUserLookup;
 }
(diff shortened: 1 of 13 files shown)


src/definition/messages/IMessage.ts
```diff
@@ -25,4 +25,8 @@ export interface IMessage {
     parseUrls?: boolean;
     customFields?: { [key: string]: any };
     blocks?: Array<IBlock>;
+    starred?: Array<{ _id: string }>;
+    pinned?: boolean;
+    pinnedAt?: Date;
+    pinnedBy?: { _id: string, username: string };
```
graywolf336, line 31: I would request that this be typed instead of generic.

src/definition/messages/IMessageFollowContext.ts
```diff
@@ -0,0 +1,21 @@
...
+     */
+    message: IMessage;
+    /**
+     * The user who follow the message
+     */
+    user: IUser;
+    /**
+     * If the message was followed or unfollowed
+     */
+    isUnfollow: boolean;
```
graywolf336, line 20: I think it would be better to be in the positive instead of being a double negative.

src/definition/messages/IMessagePinContext.ts
```diff
@@ -0,0 +1,21 @@
...
+     */
+    message: IMessage;
+    /**
+     * The user who pinned the message
+     */
+    user: IUser;
+    /**
+     * If the message was pinned or unpinned
+     */
+    isUnpinned: boolean;
```
graywolf336, line 20: Same concept here, use the positive so when it is true then it is true instead of an inverse of negative.

graywolf336 (changes requested):

src/definition/messages/IMessageFollowContext.ts
d-gubert: I agree. Even though it might seem kinda _weird_ because the name of the event is `IPostMessageFollowed`, I think the double negative will be more confusing

src/definition/messages/IMessage.ts
d-gubert: I am not so sure about this, as changing it to `IUser` would require us to add _yet another_ roundtrip to the DB to convert the message data to the app's format. This would be added to every event we listen to, and we're adding listeners to even more events. That's why I worry about the perf impact

graywolf336: I am saying something such as `IUserLookup` which would have the value that is defined here.

```ts
export interface IUserLookup {
  _id: string;
  username: string;
}
```

d-gubert: Oh I see!

d-gubert (approved):

PR #509, opened 2022-06-02
[NEW] Video Conference Read and Extend

pierre-lehnen-rc:
# What? :boat:

Created an option to read Video Conference data 

# Why? :thinking:

To use the Pexip Video Provider we'll need to register an API endpoint and this API will need to load the video conference data from the database.

# Links :earth_americas:

# PS :eyes:


thassiov (changes requested):

thassiov (approved):
LGTM

PR #521, opened 2022-06-17
[NEW] Allow apps to update setting values

pierre-lehnen-rc:
# What? :boat:

# Why? :thinking:

# Links :earth_americas:

https://github.com/RocketChat/Rocket.Chat/pull/25913

# PS :eyes:


src/server/accessors/SettingUpdater.ts
```diff
@@ -0,0 +1,20 @@
...
+        if (!this.app.getStorageItem().settings[id]) {
+            return;
+        }
+
+        const setting = await Promise.resolve(this.app.getStorageItem().settings[id]);
+
+        setting.updatedAt = new Date();
+        setting.value = value;
+
+        this.app.getStorageItem().settings[setting.id] = setting;
```
d-gubert, line 18: This method changes the setting in the current instance of the app, but since it doesn't update the database this change will not propagate to other HA instances and will be lost during a server restart.

Check this method for reference https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/improvement/vm2-poc/src/server/managers/AppSettingsManager.ts#L30

d-gubert (changes requested):

d-gubert (approved):

PR #531, opened 2022-07-04
[NEW] allow to pass the id for ephemeral messages

tapiarafael:
# What? :boat:

Added the possibility to pass an ID for ephemeral messages. This way, you could use the same ID to edit the message.

# Why? :thinking:

# Links :earth_americas:

[Rocket.chat PR](https://github.com/RocketChat/Rocket.Chat/pull/26118)

# PS :eyes:

### DIFF
--- a/src/definition/accessors/INotifier.ts
+++ b/src/definition/accessors/INotifier.ts
@@ -38,7 +38,7 @@ export interface INotifier {
      * @param user The user who should be notified
      * @param message The message with the content to notify the user about
      */
-    notifyUser(user: IUser, message: IMessage): Promise<void>;
+    notifyUser(user: IUser, message: IMessage, messageId?: string): Promise<void>;
 
     /**
      * Notifies all of the users in the provided room.
--- a/src/server/accessors/Notifier.ts
+++ b/src/server/accessors/Notifier.ts
@@ -13,14 +13,14 @@ export class Notifier implements INotifier {
         private readonly appId: string,
     ) { }
 
-    public async notifyUser(user: IUser, message: IMessage): Promise<void> {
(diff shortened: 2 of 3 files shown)


d-gubert (changes requested):
Are these changes really necessary? The `IMessage` object already has an `id` field that could be used 🤔

tapiarafael:
> 

``

tapiarafael:
> Are these changes really necessary? The `IMessage` object already has an `id` field that could be used 🤔

The MessageBuilder doesn't allow us to set the message id, and I thought it would be more explicit to pass as a separated param. But yes, we can get the message and set the id manually to avoid these changes.

d-gubert:
> > Are these changes really necessary? The `IMessage` object already has an `id` field that could be used 🤔
> 
> The MessageBuilder doesn't allow us to set the message id, and I thought it would be more explicit to pass as a separated param. But yes, we can get the message and set the id manually to avoid these changes.

Since the main objective of this is to use with ephemeral messages, and the `Notifier` receives the `IMessage` as parameter instead of a `MessageBuilder`, I think we'd be good to go by doing that

tapiarafael:
> > > Are these changes really necessary? The `IMessage` object already has an `id` field that could be used 🤔
> > 
> > 
> > The MessageBuilder doesn't allow us to set the message id, and I thought it would be more explicit to pass as a separated param. But yes, we can get the message and set the id manually to avoid these changes.
> 
> Since the main objective of this is to use with ephemeral messages, and the `Notifier` receives the `IMessage` as parameter instead of a `MessageBuilder`, I think we'd be good to go by doing that

Got it. Since these changes won't be necessary I'll close this PR.

PR #533, opened 2022-07-11
[NEW] Adding oauth apps bridge

AllanPazRibeiro:
# What? :boat:

- Added: a bridge to CRUD an OAuth app inside the rocket.chat by an app.

# Why? :thinking:

We will need this the create the Zapier app.

# Links :earth_americas:

# PS :eyes:


geekgonecrazy:
🤗 have been wanting to see this since the beginning.  Super awesome to see coming to life

src/server/AppManager.ts
```diff
@@ -878,6 +893,7 @@ export class AppManager {
         this.accessorManager.purifyApp(app.getID());
         this.uiActionButtonManager.clearAppActionButtons(app.getID());
         this.videoConfProviderManager.unregisterProviders(app.getID());
+        this.oauthAppsManager.removeAllOAuthApps(app.getID());
```
d-gubert, line 896: I'm not so sure about this change. OAuthApps are entities in Rocket.Chat, like rooms, and we don't remove all rooms created by apps when we uninstall it. I'm inclined to say we shouldn't remove it. Wdyt?

src/server/accessors/OAuthApps.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    public async getOAuthAppById(id: string, appId: string): Promise<IOAuthApp> {
+        return this.oauthAppsBridge.doGetByid(id, appId);
+    }
+
+    public async getOAuthAppByName(name: string, appId: string): Promise<Array<IOAuthApp>> {
+        return this.oauthAppsBridge.doGetByName(name, appId);
+    }
+
+    public async updateOAuthApp(OAuthApp: IOAuthApp, id: string, appId: string): Promise<void> {
```
d-gubert, line 19: Since there is an `id` property in `IOAuthApp`, can we use that instead of passing another `id` as the second paramenter here?

src/server/accessors/OAuthApps.ts
```diff
@@ -0,0 +1,26 @@
+import { IOAuthApp, IOAuthApps } from '../../definition/accessors/IOAuthApps';
+import { OAuthAppsBridge } from '../bridges/OAuthAppsBridge';
+
+export class OAuthApps implements IOAuthApps {
+    constructor(private readonly oauthAppsBridge: OAuthAppsBridge) {}
```
d-gubert, line 5: If you check the other accessors, they receive the `appId` during instantiation and keep it around to use in bridge calls. This allows them to not require an `appId` as a parameter in each of its methods

d-gubert (changes requested):
We need to resolve the previous conversations :)

src/server/accessors/OAuthApps.ts
AllanPazRibeiro: I made it like this because I was thinking about a future dependency injection that does not work when you have a mutable parameter on the constructor.

src/server/accessors/OAuthApps.ts
AllanPazRibeiro: The second id refers to the id OAuthApp that will be updated, actually, I need to change the interface of the first parameter to remove the id, since the id is unique and not mutable.

src/server/AppManager.ts
AllanPazRibeiro: I am very open on this, actually, I am not sure on this and I am ok on removing this part

src/server/managers/AppOAuthAppsManager.ts
```diff
@@ -0,0 +1,27 @@
+import { IOAuthApp, IOAuthAppParams } from '../../definition/accessors/IOAuthApps';
+import { AppManager } from '../AppManager';
+import { OAuthAppsBridge } from '../bridges/OAuthAppsBridge';
+
+export class AppOAuthAppsManager {
```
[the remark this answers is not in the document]
AllanPazRibeiro: Sure, agree I will remove this one

src/server/permissions/AppPermissions.ts
```diff
@@ -125,4 +129,6 @@ export const defaultPermissions: Array<IPermission> = [
     AppPermissions.videoConference.read,
     AppPermissions.videoConference.write,
     AppPermissions.apis.default,
+    AppPermissions['oauth-app'].write,
+    AppPermissions['oauth-app'].read,
```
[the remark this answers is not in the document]
AllanPazRibeiro: I don't get it, can you provide more details?

d-gubert (approved):

PR #534, opened 2022-07-18
Archive slashcommand

Himanshu664:
# What? :boat:

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/definition/rooms/IRoom.ts
+++ b/src/definition/rooms/IRoom.ts
@@ -1,5 +1,5 @@
-import { IUser } from '../users';
-import { RoomType } from './RoomType';
+import { IUser } from "../users";
+import { RoomType } from "./RoomType";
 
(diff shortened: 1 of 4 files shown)


src/definition/rooms/IRoom.ts
```diff
@@ -23,4 +23,5 @@ export interface IRoom {
     customFields?: { [key: string]: any };
     parentRoom?: IRoom;
     livechatData?: { [key: string]: any };
+    isArchived?: boolean;
```
debdutdeb, line 26: don't make it optional no need.

debdutdeb (changes requested):
You need to add `isArchived` to `Room` class - since it promises an IRoom. Your IDE should already let you know that along with the compile command.

PR #536, opened 2022-07-24
[NEW] Update IModifyCreator to expose IUserBuilder to enable user creation

Yuq-Bian:
# What? :boat:

Updated IModifyCreator to expose APIs to create & finish IUserBuilder to enable user creation for Roket.Chat Apps.

# Why? :thinking:

I'm a contributor for GSOC 2022 and currently working on a Rocket.Chat App to enable interop between Rocket.Chat and Microsoft Teams. The core approach for the interop is we create "dummy users" in Rocket.Chat. Each "dummy user" represents a real Teams user in Rocket.Chat world. Rocket.Chat users just need to send message to "dummy user" and the App will do the message delivery to Teams in order to build a smooth interop experience.

With this "dummy user" approach, we need this API to enable user creation for Apps.

# Links :earth_americas:

[Document for "dummy user" approach](https://github.com/RocketChat/Apps.teams.bridge/pull/4/files#diff-2095b820f1da2c36bb1743c4c8eb1fa23992b882e3cde0b6df035f55f1d58a22)

# PS :eyes:
AD-117


yash-rajpal (approved):
Looks good to me. I tested it on my local and it works fine.

graywolf336:
One suggestion when you are making "dummy users", mark them as bots. Some customers are limited in the amount of users they can have in their environments and are charged per seats, so make sure you're not accidently increasing their billing.

graywolf336:
While looking into this, we need to take into account several factors and figure out answers to some questions:

Some Rocket.Chat instances are billed per user in their system, apps shouldn't be responsible for causing huge biling. Currently users which are marked as federated are excluded from billing, so is that the route to go? Do we limit what type of users the apps can create?

Do we automatically delete users created by the app when it is uninstalled?

albuquerquefabio:
> While looking into this, we need to take into account several factors and figure out answers to some questions:
> 
> Some Rocket.Chat instances are billed per user in their system, apps shouldn't be responsible for causing huge biling. Currently users which are marked as federated are excluded from billing, so is that the route to go? Do we limit what type of users the apps can create?
> 
> Do we automatically delete users created by the app when it is uninstalled?

About the billed seats, we had a discussion and we decided to allow to creates only bot users. (It may change if necessary)

We'll implement the feature to delete all users when you uninstall the app. Probably, we'll use the **appId** to find all users created by the app and delete them.

src/definition/users/IBotUser.ts
```diff
@@ -0,0 +1,6 @@
+import { UserType } from './UserType';
+import { IUser } from './IUser';
+
+export interface IBotUser extends Omit<IUser, 'emails' | 'utcOffset'> {
```
graywolf336, line 4: Why omitting the `utcOffset`? If it is a federated user, such as teams or slack, then the app could in theory provide that value.

albuquerquefabio: Make sense!

albuquerquefabio:
AD-117

AllanPazRibeiro (approved):
LGTM

PR #539, opened 2022-08-03
Agent Override

joao-marttech:
# What? :boat:
Create Room Interface Adjustment

# Why? :thinking:
to open a room without an agent today it is necessary to perform an override, with this change it is not necessary

# Links :earth_americas:

# PS :eyes:


joao-marttech:
> # What? ⛵
> 
> Create Room Interface Adjustment
> # Why? 🤔
> 
> to open a room without an agent today it is necessary to perform an override, with this change it is not necessary
> # Links 🌎
> # PS 👀

PR #543, opened 2022-08-19
Chore: Adjustments for using raw models at rocket chat

thassiov:
# What? :boat:

Adjustments needed at the main repo for the migration to service architecture.

# Why? :thinking:

_This is part of the initiative to make the apps engine work as a service._
# Links :earth_americas:

https://github.com/RocketChat/Rocket.Chat/pull/26633

# PS :eyes:


src/definition/persistence/IPersistenceItem.ts
```diff
@@ -0,0 +1,7 @@
+import { RocketChatAssociationRecord } from '../../definition/metadata';
+
+export interface IPersistenceItem {
+    appId: string;
+    data: { [key: string]: object };
```
d-gubert, line 5:
```suggestion
    data: Record<string, any>;
```

src/server/storage/AppMetadataStorage.ts
```diff
@@ -8,7 +8,7 @@ export abstract class AppMetadataStorage {
     }
 
     public abstract create(item: IAppStorageItem): Promise<IAppStorageItem>;
-    public abstract retrieveOne(id: string): Promise<IAppStorageItem>;
+    public abstract retrieveOne(id: string): Promise<IAppStorageItem | null>;
```
d-gubert, line 11: This looks like a breaking change, right? If so, we'd need to leave it to v2.0

src/definition/persistence/IPersistenceItem.ts
```diff
@@ -0,0 +1,7 @@
...
+    associations?: Array<RocketChatAssociationRecord>;
```
d-gubert, line 6: It seems weird to me for the `associations` field to be optional 👀 can you please confirm?

d-gubert (changes requested):

src/definition/persistence/IPersistenceItem.ts
thassiov: Yes, it should be optional to conform to our persistence model. We have methods to CRUD an item without the `associations` field ([example](https://github.com/RocketChat/Rocket.Chat/blob/f2209eecda95b7a59f04b73e14a79cae9be2dd05/apps/meteor/app/apps/server/bridges/persistence.ts#L18)).

d-gubert: That method is definitely weird, but let's leave this investigation for a better moment

src/definition/persistence/IPersistenceItem.ts
```diff
@@ -0,0 +1,7 @@
+import { RocketChatAssociationRecord } from '../../definition/metadata';
+
+export interface IPersistenceItem {
+    appId: string;
+    data: Record<string, unknown>;
+    associations: Array<RocketChatAssociationRecord>;
```
d-gubert, line 6:
```suggestion
    associations?: Array<RocketChatAssociationRecord>;
```

d-gubert (approved):

PR #546, opened 2022-08-30
[FIX] scheduler.cancelJob() did not return the proper Promise

murtaza98:
# What? :boat:

# Why? :thinking:

if we either use `cancelJob` or `cancelAllJobs`  method of `ISchedulerModify`, the operation runs asynchronously due to which there is no way for apps to know if the operation was completed. This PR attempts to fix this

Sample code:
```
await configurationModify.scheduler.cancelJob(
				JobId.JOBS_RESTART_JOB,
);
```
Note: even with the `await` above, it still runs asynchronously  because the promise is not returned back to the app from the code :grimacing: 
# Links :earth_americas:

# PS :eyes:


debdutdeb (approved):
lgtm :)

murtaza98:
Hi @RocketChat/apps Team!  This issue still persists in the current version of the apps-engine. Can you please check this PR? :pray: Thanks in advance :hugs:

d-gubert:
Hey @murtaza98 ! Thanks for the fix. Just a quick clarification: the problem was that the cancel methods were not returning the Promise created on the bridges, and thus it was resolving more quickly than the correct promise. But the operation is always (and will continue being) asynchronous.

PR #550, opened 2022-09-13
[FEAT] Assign ids to blocks and elements

tapiarafael:
# What? :boat:

Assign blockId, appId, and actionId for the blocks and elements when calling openSurfaceView/updateSurfaceView method and on finish message builder.

# Why? :thinking:

# Links :earth_americas:

[Task AE-55](https://rocketchat.atlassian.net/jira/software/c/projects/AE/boards/28?modal=detail&selectedIssue=AE-55)

# PS :eyes:

### DIFF
--- a/src/definition/package.json
+++ b/src/definition/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@rocket.chat/apps-ts-definition",
-  "version": "1.33.0-alpha",
+  "version": "1.35.0-alpha",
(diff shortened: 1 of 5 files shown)


thassiov (changes requested):
Looks fine, but I would suggest putting the `_assignIds` method somewhere inside the [`src/server/misc`](https://github.com/RocketChat/Rocket.Chat.Apps-engine/tree/alpha/src/server/misc) directory so it can be shared between modules. WDYT?

thassiov (approved):

PR #554, opened 2022-11-12
[New] Support for x-www-form-urlencoded Body Params in OAuth2 calls

mustafahasankhan:
# What? :boat:

- Added support for OAuth2 Clients that require parameters in Body as x-www-form-urlencoded Content-Type.
- Backwards compatible, does not affect client that don't require parameters in Body.

# Why? :thinking:

We need this in using Microsoft Graph API and other OAuth2 Clients.

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/server/oauth2/OAuth2Client.ts
+++ b/src/server/oauth2/OAuth2Client.ts
@@ -76,16 +76,6 @@ export class OAuth2Client implements IOAuth2Client {
                 packageValue: '',
                 i18nLabel: `${this.config.alias}-oauth-client-secret`,
             }),
-
-            configuration.settings.provideSetting({
-                id: `${this.config.alias}-request-content-type`,
-                type: SettingType.STRING,
-                public: true,
-                required: false,
-                packageValue: '',
-                values: [{key:"0", i18nLabel:"No"},{key:"1", i18nLabel:"Yes"}],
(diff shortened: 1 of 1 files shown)


src/server/oauth2/OAuth2Client.ts
```diff
@@ -30,8 +30,8 @@ export class OAuth2Client implements IOAuth2Client {
     private defaultContents = {
         success:  '<div style="display: flex;align-items: center;justify-content: center; height: 100%;">\
                         <h1 style="text-align: center; font-family: Helvetica Neue;">\
-                            Authorization went successfully<br>\
-                            You can close this tab now<br>\
+                            Authorization successful.<br>\
```
AllanPazRibeiro, line 33: I will ask you to let this message as it was previously

AllanPazRibeiro (changes requested):

AllanPazRibeiro (dismissed):

thassiov (dismissed):

AllanPazRibeiro (approved):

mustafahasankhan:
Leaving this comment for the CLA-assistant so that it refreshes the CLA check for this PR.

d-gubert:
I'll merge this. The CLA is complaining about the two first commits that haven't been made with the correct configuration, and fixing that would be a hassle, so we'll just ignore it in this case

PR #565, opened 2022-12-22
[FEAT] Add checksum and signature to apps

tapiarafael:
# What? :boat:

Add a checksum to validate apps information and make a signature

# Why? :thinking:

# Links :earth_americas:

[AE-70](https://rocketchat.atlassian.net/browse/AE-70?atlOrigin=eyJpIjoiYzdhNGI2MzE5NzYxNDMwYjkxZjQ4M2U3OWNjMzVhYmQiLCJwIjoiaiJ9)

# PS :eyes:
Depends on [PR](https://github.com/RocketChat/Rocket.Chat/pull/27615)

[AE-70]: https://rocketchat.atlassian.net/browse/AE-70?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/server/AppManager.ts
```diff
@@ -1032,6 +1041,30 @@ export class AppManager {
...
+    private stringifyOrdered(obj: any): string {
+        return JSON.stringify(obj, Object.keys(obj).sort());
+    }
+
+    private async getAppSignature(obj: any, appId: string): Promise<string> {
+        const descriptorString = this.stringifyOrdered(obj);
+        const calg = 'SHA256';
+        const alg = 'RS512';
+        const checksum = this.checksum(descriptorString, calg);
+        const pkeyString = await AppManager.Instance.getBridges().getInternalFederationBridge().getPrivateKey();
```
d-gubert, line 1058: Curious to why you needed the `AppManager.Instance` part if you are inside the `AppManager` class 🤔

src/server/AppManager.ts
```diff
@@ -1032,6 +1041,30 @@ export class AppManager {
 
         return result;
     }
+
+    private checksum(str: string, alg = 'SHA256'): string {
+        return createHash(alg).update(str).digest('hex');
+    }
+
+    private stringifyOrdered(obj: any): string {
+        return JSON.stringify(obj, Object.keys(obj).sort());
```
d-gubert, line 1050: On the app update flow, this will cause the descriptor string to include the signature itself to be hashed. I'm wondering if this won't cause problems since we'll be updating that signature after calculating the hash.

d-gubert (changes requested):

src/server/AppManager.ts
d-gubert: Maybe this would be enough?

```suggestion
        return JSON.stringify(obj, Object.keys(obj).filter(key => key !== 'signature').sort());
```

src/server/AppManager.ts
```diff
@@ -601,6 +603,8 @@ export class AppManager {
             return aff;
         }
 
+        const signature = await this.getAppSignature(descriptor, result.info.id);
+        descriptor.signature = signature;
```
graywolf336, line 607: Any particular reason to await to a variable to only then assign that variable to a property on an object? Why not the following:
```suggestion
        descriptor.signature = await this.getAppSignature(descriptor, result.info.id);
```

d-gubert (approved):

PR #577, opened 2023-01-19
[NEW] Add "password" setting type

pierre-lehnen-rc:
# What? :boat:

Adds "Password" to the list of allowed setting types

# Why? :thinking:
So apps can create password settings.

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/definition/settings/SettingType.ts
+++ b/src/definition/settings/SettingType.ts
@@ -7,5 +7,7 @@ export enum SettingType {
     SELECT = 'select',
     STRING = 'string',
     MULTI_SELECT = 'multiSelect',
+	// Renders an input of type 'password' in the form. IMPORTANT - the value will NOT be encrypted
+	// it will be treated as a password just on the screen
     PASSWORD = 'password',
 }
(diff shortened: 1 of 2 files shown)


src/definition/settings/SettingType.ts
```diff
@@ -7,4 +7,5 @@ export enum SettingType {
     SELECT = 'select',
     STRING = 'string',
     MULTI_SELECT = 'multiSelect',
+    PASSWORD = 'password',
```
graywolf336, line 10: We probably need to add documentation to this to state it doesn’t encrypt the value or anything like that.

d-gubert:
```suggestion
	// Renders an input of type 'password' in the form. IMPORTANT - the value will NOT be encrypted
	// it will be treated as a password just on the screen
    PASSWORD = 'password',
```

src/definition/settings/SettingType.ts
```diff
@@ -7,4 +7,7 @@ export enum SettingType {
     SELECT = 'select',
     STRING = 'string',
     MULTI_SELECT = 'multiSelect',
+	// Renders an input of type 'password' in the form. IMPORTANT - the value will NOT be encrypted
+	// it will be treated as a password just on the screen
+    PASSWORD = 'password',
```
d-gubert, line 12:
```suggestion
    // Renders an input of type 'password' in the form. IMPORTANT - the value will NOT be encrypted
    // it will be treated as a password just on the screen
    PASSWORD = 'password',
```

d-gubert (approved):

Issue #238, opened 2020-03-17
Add masked SettingType


apazureck:
Hi,

I would like to write an app where a username and a password is required.

Is there any (undocumented) option to create a masked text box in the settings for a custom app?

For now I use `SettingType.STRING`, but then the password can be read and easily copied. It would be nice if there was a masked / password text box the user can put their password in.

Cheers!

apazureck:
Would this be a good issue to start? If it is not that hard I would try to get it done. Where could I start?

PR #582, opened 2023-02-04
Fixed Masked SettingType

ashnove:
This closes https://github.com/RocketChat/Rocket.Chat.Apps-engine/issues/238

The fix/hack I've added is to append a PASSWORD attribute to the SettingType enum and used it in the provideSetting configuration.


graywolf336:
Is this a duplicate of #577?

ashnove:
> Is this a duplicate of #577?

Seems like it

pierre-lehnen-rc:
Just the first part - My PR only adds password to the list, but doesn't change any setting to use it.

PR #591, opened 2023-02-16
[FIX] Remove settings from checksum fields

tapiarafael:
# What? :boat:

Remove settings from the checksum validation.

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:


d-gubert (approved):
vamoooooooo

PR #637, opened 2023-06-06
feat:  Thread Bridge

ggazzo:
# What? :boat:

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:


src/definition/accessors/IThreadRead.ts
```diff
@@ -0,0 +1,9 @@
+import { IMessage } from '../messages/index';
+
+/**
+ * This accessor provides methods for accessing
+ * Thread messages in a read-only-fashion.
+ */
+export interface IThreadRead {
+    getThreadById(id: string): Promise<Array<IMessage> | undefined>;
```
d-gubert, line 8: I guess this opens up that possibility of clogging up the server process if the app requests a _very big thread_ :tm: This is one of the reasons we've been refraining from adding this kind of API to the engine without better filter options and etc.

What is your opinion on this @ggazzo ?

ggazzo: I think the query executed and the impact will be the same as when we open a thread and make a rest request to get the list. I don't see how this could destroy the ecosystem

ggazzo: maybe instead of IMessage we could return something with less fields, at least 'id' | 'msg' | 'u' would be ok

d-gubert: > I don't see how this could destroy the ecosystem

By loading potentially hundreds of thousands of messages per query, which can be executed several times in a short period. Granted, it is _less likely_ for threads to grow and become as big as a very old room, for instance

> make a rest request to get the list

I was considering that this rest request always paginated the query. If that's not the case and this is not a concern, then we should be good

ggazzo: we don't page(I think we should, but because of the frontend, if the app is bad and causes problems, we can disable it, the frontend can't be turned off) threads today and it hasn't happened to be a problem yet, even though several people opening a room at the same time would offer the same risk.

what would happen if the app decided to fetch all the pages and get all the data anyway? the same result

maybe we should introduce some debounce or bulkhead mechanism, but then this should apply to all methods in the app

d-gubert (approved):

PR #638, opened 2023-06-07
feat: Extend typing capability for threads

ggazzo:
# What? :boat:

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:


d-gubert (changes requested):
Unfortunately this is a breaking change, since the type `TypingScope` has been removed and the parameter for the `typing` function has changed. It will cause developers who try to update their apps with the latest compatible version to face an unexpected error while packaging the app. For these changes to go forward as they are, it needs to go in a major update of the Framework or be adapted to be compatible with the current type definitions

PR #645, opened 2023-07-07
Updated: Livechat-Room.visitor definitions to include lastMessageTs prop

murtaza98:
# What? :boat:

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/definition/livechat/ILivechatRoom.ts
+++ b/src/definition/livechat/ILivechatRoom.ts
@@ -29,7 +29,7 @@ type OmnichannelSource =
     | IOmnichannelSourceApp;
 
 export interface ILivechatRoom extends IRoom {
-    visitor: IVisitor;
+    visitor: IVisitor & { lastMessageTs?: Date };
     department?: IDepartment;
     servedBy?: IUser;
     responseBy?: IUser;


src/definition/livechat/ILivechatRoom.ts
```diff
@@ -29,7 +29,7 @@ type OmnichannelSource =
     | IOmnichannelSourceApp;
 
 export interface ILivechatRoom extends IRoom {
-    visitor: IVisitor;
+    visitor: IVisitor & { lastMessageTs?: Date };
```
d-gubert, line 32: I'm a bit confused on why you've introduced the new property with the intersection operator here instead of in `IVisitor` itself, can you clarify?

d-gubert (changes requested):
I'm a bit confused on why

src/definition/livechat/ILivechatRoom.ts
murtaza98: Hi @d-gubert I share more context about this over [here](https://github.com/RocketChat/Rocket.Chat/pull/29756/files#r1257265249). 
In-short: this additional prop is stored on room level since we can have a single visitor contacting us using multiple channels, hence we're storing all channel related info on room level. LastMessageTs is one of such channel level prop.

d-gubert (changes requested):
Create the new field as discussed in the conversation above

src/definition/livechat/ILivechatRoom.ts
murtaza98: Thanks for the feedback Doug. I've updated the PR with your suggestions

PR #658, opened 2023-08-25
Add new variant prop to ActionButton types

csuadev:
# What? :boat:
- Added type to suppor danger variant inside ActionButtons. 

# Why? :thinking:

# Links :earth_americas:
[https://rocketchat.atlassian.net/browse/AECO-97](https://rocketchat.atlassian.net/browse/AECO-97)

# PS :eyes:

### DIFF
--- a/src/definition/ui/IUIActionButtonDescriptor.ts
+++ b/src/definition/ui/IUIActionButtonDescriptor.ts
@@ -1,4 +1,3 @@
-import type { Keys as IconName } from '@rocket.chat/icons';
 import { UIActionButtonContext } from './UIActionButtonContext';
 
 export enum RoomTypeFilter {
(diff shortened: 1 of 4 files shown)


src/definition/ui/IUIActionButtonDescriptor.ts
```diff
@@ -30,8 +30,9 @@ export interface IUActionButtonWhen {
 
 export interface IUIActionButtonDescriptor {
     actionId: string;
-    labelI18n: string;
     context: UIActionButtonContext;
+    labelI18n: string;
+    variant?: 'danger';
```
d-gubert, line 35: Just clarifying, this means that the `variant` prop can either be omitted in the descriptor or be set to `'danger'`, is that correct? I'm thinking if it wouldn't be worth it to add a `'normal'` variant - it feels more ergonomic for situations where I need to dynamically assign the danger vairant, where I can do like `variant: isDanger ? 'danger' : 'normal'`, instead of `...isDanger && { variant: 'danger' }`

d-gubert (approved):

PR #659, opened 2023-08-30
chore(deps): Bump a bunch of development dependencies

tassoevan:
# What? :boat:

- Upgraded `uuid` dependency (and `@types/uuid`) to match the main repository;
- Upgraded `typescript` and `typedoc`;
- Replaced `tslint` with  `eslint`.

# Why? :thinking:

It was reported that some ocasional collisions between types of `uuid` exist, therefore is not far-fetched to assume there is a risk of colliding the library itself.
Also, using the standard ESLint configuration would keep the consistency we adhere in other codebases in the organization.

# Links :earth_americas:

# PS :eyes:


d-gubert (approved):
Looks legit, thank you so much!

PR #660, opened 2023-09-01
feat: add dm notify method

Dnouv:
# What? :boat:

- Added a new helper to notify the user via a DM from the App.
# Why? :thinking:

- Easier way of DM'ing users without managing custom logic inside their App.

# Links :earth_americas:

[AECO-360](https://rocketchat.atlassian.net/browse/AECO-360)
https://github.com/RocketChat/Rocket.Chat/pull/30321

# PS :eyes:
This does not have a dependent PR on the Core; the introduced method utilizes the pre-existing set of room and message helpers to get the task done.

[AECO-360]: https://rocketchat.atlassian.net/browse/AECO-360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/server/managers/AppAccessorManager.spec.ts
```diff
@@ -127,7 +127,7 @@ export class AppAccessorManagerTestFixture {
         Expect(this.bridges.getServerSettingBridge).toHaveBeenCalled().exactly(1);
         Expect(this.bridges.getEnvironmentalVariableBridge).toHaveBeenCalled().exactly(1);
         Expect(this.bridges.getPersistenceBridge).toHaveBeenCalled().exactly(1);
-        Expect(this.bridges.getRoomBridge).toHaveBeenCalled().exactly(1);
+        Expect(this.bridges.getRoomBridge).toHaveBeenCalled().exactly(2);
```
AlenDavid, line 130: Is it really needed to be called twice? Can't we reuse the first return value from getRoomBridge on both cases?

src/server/accessors/Notifier.ts
```diff
@@ -46,4 +71,27 @@ export class Notifier implements INotifier {
...
+                usernames: [user.username, sender.username],
+                creator: sender,
+                id: '',
+                slugifiedName: '',
+            },
+            [user.username, sender.username],
+            this.appId,
+        );
+
+        return this.roomBridge.doGetById(newDMRId, this.appId);
```
debdutdeb, line 95: Don't need it, you can save this call. 

The bridge already catches the room object by passed id. So on L51, you can simply do `room: { id: newId }` and here 
```suggestion
        return newDMRId;
```

Btw, typo in `RId` :)

src/server/accessors/Notifier.ts
```diff
@@ -46,4 +71,27 @@ export class Notifier implements INotifier {
...
+
+    /**
+     * Creates a new direct message room between two users.
+     *
+     * @param {IUser} user - The first user to create the room for.
+     * @param {IUser} sender - The second user to create the room for.
+     * @returns {Promise<IRoom>} A Promise that resolves with the newly created direct message room.
+     */
+    private async createDMRoom(user: IUser, sender: IUser): Promise<IRoom> {
+        const newDMRId = await this.roomBridge.doCreate(
```
debdutdeb, line 83: you have a createDirectMessage method available in room bridge iirc

debdutdeb: but maybe that's hidden behind create, don't remember exactly.

src/server/accessors/Notifier.ts
```diff
@@ -46,4 +71,27 @@ export class Notifier implements INotifier {
...
+            {
+                type: RoomType.DIRECT_MESSAGE,
+                usernames: [user.username, sender.username],
```
debdutdeb, line 86: dont think is required/used

src/server/accessors/Notifier.ts
```diff
@@ -46,4 +71,27 @@ export class Notifier implements INotifier {
...
+                creator: sender,
+                id: '',
+                slugifiedName: '',
```
debdutdeb, line 89: Same here

src/server/accessors/Notifier.ts
```diff
@@ -30,6 +35,26 @@ export class Notifier implements INotifier {
...
 
+    /**
+     * Sends a direct message to a user.
+     *
+     * @param {IUser} user - The user to send the message to.
+     * @param {IMessage} partialMsg - The partial message to send.
+     * @returns {Promise<void>} A Promise that resolves when the message has been sent.
+     */
+    public async notifyDM(user: IUser, partialMsg: Omit<IMessage, 'room'>): Promise<void> {
+        const sender = partialMsg.sender || (await this.userBridge.doGetAppUser(this.appId));
```
debdutdeb, line 46: Do we anywhere else hve this default behaviour of falling back to app user? if not this would be inconsistent and confusing IMO. In that case method should fail if sender is not passed (iirc sender is mandatory in Imessage)

debdutdeb (changes requested):
my one last comment would be on the naming. Full descriptive names are better than abbreviations IMO, thus `notifyDirectMessage` or `notifyViaDirectMessage` would be better than `DM`.

Open to other POVs, as another consideration is to be cosistent with the naming conventions in other places, where I fall short of knowledge.

src/server/accessors/Notifier.ts
```diff
@@ -30,6 +35,26 @@ export class Notifier implements INotifier {
...
+        const dmRoom = (await this.roomBridge.doGetDirectByUsernames([user.username, sender.username], this.appId)) || (await this.createDMRoom(user, sender));
+
+        const message: IMessage = {
+            ...partialMsg,
+            room: dmRoom,
+            sender,
+        };
+
+        await this.msgBridge.doCreate(message, this.appId);
```
debdutdeb, line 55: also, should this not call doNotifyRoom ? Notifier is only for ephemeral messages and doCreate would create a normal message ..

Dnouv: Thank you for the review, Deb! Actually, no, this is a method to be used by the Apps to notify users via a DM.

src/server/accessors/Notifier.ts
Dnouv: It is required, or TS errors pop up. I think this is planned to be removed in `2.0.0`

src/server/accessors/Notifier.ts
Dnouv: It is required, or TS errors pop up.

src/server/accessors/Notifier.ts
Dnouv: Oh, no, there isn't any `createDirectMessage` method in the room bridge, but there is a meteor method with that name, which is used inside the `create` (on the Apps.Engine calls it using the `doCreate`) room bridge.
https://github.com/RocketChat/Rocket.Chat/blob/ae5f958712cb31640fa704756382f6b7b061dc00/apps/meteor/app/apps/server/bridges/rooms.ts#L32-L33

src/server/accessors/Notifier.ts
Dnouv: My bad, I will correct the `RId` 🙈.

> The bridge already catches the room object by passed id. So on L51, you can simply do room: { id: newId } and here

Haha, actually, it will give a TS error as well,
```
Type '{ id: string; }' is missing the following properties from type 'IRoom': slugifiedName, type, creator, usernames
```

src/server/accessors/Notifier.ts
debdutdeb: okay but the thing is all notifier methods are ephemeral messages and that's what I, at least associate the accessor to function as.

At least if this has to be under notifier accessor, could the method name not use the `notify` keyword as to distinguish between ephemeral and persistent messages?

I'm saying this as a potential consumer of this api, that it can be confusing.

src/server/accessors/Notifier.ts
debdutdeb: 🤦🏼‍♂️ damn you typescript sometimes ...

could you add a small comment to clear that up? Might feel unnecessary but when reading the code my first thought was 'what is the significance of passing empty strings here' - while there wasn't any

src/server/accessors/Notifier.ts
debdutdeb: Same as the last comment .. but maybe also pass an empty array instead as it's not used anyway.

src/server/accessors/Notifier.ts
debdutdeb: ts errors in this place is IMO unhelpful rather even bad. to satisfy the linter/compiler, we're adding runtime weight for no benefit.

I'd force cast it. 

But, maybe update the bridge method to just accept the id instead. Mostly because if anybody is editing the bridge code, sees IRoom being passed, can just take all its fields for granted and start consuming them in the future, breaking usages like a force cast.

Dnouv: Thank you, Deb! Sure, I will check and modify it. I will also check if there aren't any dependents for these fields.

src/server/accessors/Notifier.ts
Dnouv: Ah, interesting, but won't the `notifyDirectMessage` already explain that it will be a DM? I mean, if we remove the `notify`, it might create further confusion because it was designed to notify the user via a DM, or else there would be a thought: _Does it perhaps a new helper to let me easily create DMs_. And this might lead to missing the actual use-case. 

Do we need to update perhaps the `jsDoc`? I tried to keep it as informative as possible, but please do let me know if it is not able to deliver the correct message. I would appreciate it.

Thank you!

src/server/accessors/Notifier.ts
debdutdeb: Don't worry about it, just trying to ease up our own code jumps when reading in the future :)

tests/server/managers/AppAccessorManager.spec.ts
AlenDavid: Thanks for the clarification!

src/server/accessors/Notifier.ts
Dnouv: I have added the comments but did not make any changes to the methods param types because, after looking into the converters (https://github.com/RocketChat/Rocket.Chat/blob/ae5f958712cb31640fa704756382f6b7b061dc00/apps/meteor/app/apps/server/converters/rooms.js#L81) the `slugifiedName` is used as the room name, and if I make it optional while creating the channels, it will be problematic.

A possible solution would be to separate the DM room creation, but that's a bit strange because the way we have `create` now with `switch` logic looks good.

AllanPazRibeiro:
Perhaps the most appropriate location for this method could be within the IModify accessor. Additionally, the current method name doesn't quite align with its purpose, as all notifications are ephemerals. Consider renaming it to something more straightforward, like `sendDirectMessage`, instead of the somewhat cryptic `notifyDM`. Is there a specific requirement for this method to be included in the IRead interface?

AllanPazRibeiro (changes requested):
Perhaps the most appropriate location for this method could be within the IModify accessor. Additionally, the current method name doesn't quite align with its purpose, as all notifications are ephemerals. Consider renaming it to something more straightforward, like `sendDirectMessage`, instead of the somewhat cryptic `notifyDM`. Is there a specific requirement for this method to be included in the IRead interface?

AllanPazRibeiro:
> Hey @AllanPazRibeiro 
> 
> Thank you for the feedback. Yes, I think that would be a good name change; I will change it. 
> 
> About keeping it inside the `IRead`, I plan to introduce it as a way to allow Apps to notify users via a DM. And for the `IPre...` events, there is no `IModify`. I have added more details on, the ADR, -> https://adr.rocket.chat/62
> 
> Thank you!

Makes senses to me!

AllanPazRibeiro (dismissed):
LGTM

src/server/accessors/Notifier.ts
debdutdeb: > A possible solution would be to separate the DM room creation, but that's a bit strange because the way we have create now with switch logic looks good.

I like this. I don't understand why it'd be strange to do so. If typescript is the problem here, then a union type can fix it rather easily `{ type: RoomType.DIRECT_MESSAGE; id: string} | IRoom` - but personally i find this weird, while being a quick fix.

debdutdeb: the union should work .. let me see in playground

debdutdeb: https://www.typescriptlang.org/play?#code/KYOwrgtgBASg9nCAVAngB2FA3gKCvqAEQEkYBRAYSQH0BZMgZQYEEBxMqAXigHIATCDwA0eAgAUAqgCEAMsQpdeaMACNho-GJjEAasyQduPNACcAljxwBfHDjMgALsBMAzAIYBjTKQTRcBKAd0YAAuWF9UDABuDSgzPjCAZwdzEABzGIDEgBswNKSU+zTrWxcwEA8HMzgQKBcEAAo0NxRsuDcE7CCMMPhESOAAOhJyKjpGFnYouM7k1LSrKAAfKB9EAEpsWIB6bZKcergG-wJu0PD+4OHSShp6JjYyEQD4sJ5LK3XSxpP8M96IlctLp9E9Yq9eB8vjggA

debdutdeb: > A possible solution would be to separate the DM room creation, but that's a bit strange because the way we have create now with switch logic looks good.

I like this. I don't understand why it'd be strange to do so. 

If typescript is the problem here, then a union type can fix it rather easily `{ type: RoomType.DIRECT_MESSAGE; id: string} | IRoom` - but personally i find this weird, while being a quick fix. The converter won't be a problem as the resultants will just be undefined (but maybe check)

debdutdeb: union should work how im thinking .. ill check

debdutdeb: https://www.typescriptlang.org/play?#code/KYOwrgtgBASg9nCAVAngB2FA3gKCvqAEQEkYBRAYSQH0BZMgZQYEEBxMqAXigHIATCDwA0eAgAUAqgCEAMsQpdeaMACNho-GJjEAasyQduPNACcAljxwBfHDjMgALsBMAzAIYBjTKQTRcBKAd0YAAuWF9UDABuDSgzPjCAZwdzEABzGIDEgBswNKSU+zTrWxcwEA8HMzgQKBcEAAo0NxRsuDcE7CCMMPhESOAAOhJyKjpGFnYouM7k1LSrKAAfKB9EAEpsWIB6bZKcergG-wJu0PD+4OHSShp6JjYyEQD4sJ5LK3XSxpP8M96IlctLp9E9Yq9eB8vjggA

debdutdeb (approved):
Thanks for making the changes :)

AllanPazRibeiro (approved):
LGTM

PR #668, opened 2023-10-30
Create mocks to call accessors from Deno

d-gubert:
# What? :boat:

We need to create accessor mocks in Deno that will call the Node process.

The mocks can be proxies that will issue a Request message to the Node process that will actually perform the method call in the appropriate accessor, returning a Promise that will resolve when Node sends a Response message to the initiating Request.

On the Apps-Engine runtime, this request will be handled by the DenoRuntimeSubprocessController instance associated with the originating subprocess
# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/deno-runtime/deno.jsonc
+++ b/deno-runtime/deno.jsonc
@@ -1,5 +1,6 @@
 {
     "imports": {
+        "@rocket.chat/apps-engine/": "./../src/",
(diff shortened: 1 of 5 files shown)


d-gubert:
Closing in favor of #670 670

PR #684, opened 2023-12-08
Refactor app initialization

d-gubert:
# What? :boat:

Refactor app initialization with new runtime
# Why? :thinking:

# Links :earth_americas:

# PS :eyes:


src/server/managers/AppVideoConfProviderManager.ts
```diff
@@ -100,7 +100,9 @@ export class AppVideoConfProviderManager {
     }
 
     public async isFullyConfigured(providerName: string): Promise<boolean> {
+        console.log(1)
         const providerInfo = this.retrieveProviderInfo(providerName);
+        console.log(2)
```
d-gubert, line 105:
```suggestion
        const providerInfo = this.retrieveProviderInfo(providerName);
```

deno-runtime/main.ts
```diff
@@ -54,13 +53,19 @@ function wrapAppCode(code: string): (require: (module: string) => unknown) => Pr
...
     const exports = await wrapAppCode(source)(require);
+
     // This is the same naive logic we've been using in the App Compiler
-    const appClass = Object.values(exports)[0] as typeof App;
-    const app = new appClass({ author: {} }, console, AppAccessorsInstance.getDefaultAppAccessors());
+    // Applying the correct type here is quite difficult because of the dynamic nature of the code
+    // deno-lint-ignore no-explicit-any
+    const appClass = Object.values(exports)[0] as any;
+
+    // What do I do here? D:
```
d-gubert, line 67:
```suggestion
    // What do I do here? D:
```

Cast to `any`, that's what we can do here

PR #689, opened 2023-12-22
Prevent proxies from sending toJSON calls via IPC

d-gubert:
# What? :boat:

The proxies used to send method calls as request over IPC ended up also sending calls to the `toJSON` method as a request. This, of course, makes no sense and only causes problems.

This PR also adds a `toJSON` implementation to the `App` abstract class, as the app instance can be sent over for configuration methods if the object sent has a reference to the App instance. E.g. instantiating a slashcommand class that takes the instace of the App as parameter via constructor.
# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/definition/App.ts
+++ b/src/definition/App.ts
@@ -220,4 +220,9 @@ export abstract class App implements IApp {
         this.logger.debug(`The status is now: ${status}`);
         this.status = status;
     }
+
+    // Avoid leaking references if object is logged or sent over IPC
+    public toJSON(): Record<string, any> {
+        return this.info;
+    }
 }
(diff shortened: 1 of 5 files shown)


src/definition/App.ts
```diff
@@ -220,4 +220,9 @@ export abstract class App implements IApp {
         this.logger.debug(`The status is now: ${status}`);
         this.status = status;
     }
+
+    // Avoid leaking references if object is logged or sent over IPC
```
d-gubert, line 224:
```suggestion
    // Avoid leaking references if object is serialized (e.g. to be sent over IPC)
```

PR #704, opened 2024-01-12
Improve incoming message handling on both Node and Deno

d-gubert:
# What? :boat:

Both ends of the socket now can receive multiple messages or incomplete messages and will handle this appropriately
# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/server/runtime/AppsEngineDenoRuntime.ts
+++ b/src/server/runtime/AppsEngineDenoRuntime.ts
@@ -132,12 +132,16 @@ export class DenoRuntimeSubprocessController extends EventEmitter {
         }
     }
 
+    private send(message: jsonrpc.JsonRpc) {
+        this.deno.stdin.write(message.serialize().concat(MESSAGE_SEPARATOR));
+    }
+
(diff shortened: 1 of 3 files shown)


src/server/runtime/AppsEngineDenoRuntime.ts
```diff
@@ -179,10 +183,58 @@ export class DenoRuntimeSubprocessController extends EventEmitter {
...
 
     private setupListeners(): void {
-        this.deno.stdout.on('data', this.parseOutput.bind(this));
         this.deno.stderr.on('data', this.parseError.bind(this));
-
         this.on('ready', this.onReady.bind(this));
+
+        let messageBuffer: string[] = [];
+
+        this.deno.stdout.on('data', async (chunk: Buffer) => {
```
d-gubert, line 191: I had to move this out because I needed the `messageBuffer` to live outside the scope of the callback (previously `parseOutput`).

There is likely a better way to do this, but it can be left to another time maybe

PR #705, opened 2024-01-15
fix: add Buffer conversion for file uploads

tapiarafael:
# What? :boat:

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/deno-runtime/handlers/app/construct.ts
+++ b/deno-runtime/handlers/app/construct.ts
@@ -31,11 +31,12 @@ function wrapAppCode(code: string): (require: (module: string) => unknown) => Pr
     return new Function(
         'require',
         `
+        const Buffer = require('node:buffer');
         const exports = {};
         const module = { exports };
-        const result = (async (exports,module,require,globalThis,Deno) => {
+        const result = (async (exports,module,Buffer,require,globalThis) => {
             ${code};
-        })(exports,module,require);
+        })(exports,module,require,Buffer);
         return result.then(() => module.exports);`,
     ) as (require: (module: string) => unknown) => Promise<Record<string, unknown>>;
 }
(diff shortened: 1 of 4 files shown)


deno-runtime/handlers/app/construct.ts
```diff
@@ -31,11 +31,12 @@ function wrapAppCode(code: string): (require: (module: string) => unknown) => Pr
     return new Function(
         'require',
         `
+        const Buffer = require('node:buffer');
         const exports = {};
         const module = { exports };
-        const result = (async (exports,module,require,globalThis,Deno) => {
+        const result = (async (exports,module,Buffer,require,globalThis) => {
```
d-gubert, line 37:
```suggestion
        const result = (async (exports,module,require,Buffer,globalThis, Deno) => {
```

The order of parameters in the call is different, and we should still "shadow" the `Deno` name so it's not so easily accessible

deno-runtime/lib/accessors/mod.ts
```diff
@@ -189,7 +193,23 @@ export class AppAccessors {
...
+                    },
+                    // Convert the Uint8Array to a Buffer 
+                    getBufferById(id: string) {
+                        const result = this._proxy.getBufferById(id);
+                        return Buffer.from(result);
+                    },
+                    // Convert the Uint8Array to a Buffer 
+                    getBuffer(upload: IUpload) {
+                        const result = this._proxy.getBuffer(upload);
+                        return Buffer.from(result);
```
d-gubert, line 209:
```suggestion
                    async getBufferById(id: string) {
                        const result = await this._proxy.getBufferById(id);
                        return Buffer.from(result);
                    },
                    // Convert the Uint8Array to a Buffer 
                    async getBuffer(upload: IUpload) {
                        const result = await this._proxy.getBuffer(upload);
                        return Buffer.from(result);
```

Calls to proxy methods always return a promise

d-gubert (approved):

PR #707, opened 2024-01-20
Fix batch

d-gubert:
# What? :boat:

- Register response listener before sending the message
- Fix imports in slashcommand file

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/deno-runtime/lib/messenger.ts
+++ b/deno-runtime/lib/messenger.ts
@@ -144,8 +144,6 @@ export async function sendRequest(requestDescriptor: RequestDescriptor): Promise
 
     await Transport.send(request);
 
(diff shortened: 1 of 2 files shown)


deno-runtime/lib/messenger.ts
```diff
@@ -143,6 +141,12 @@ export async function sendRequest(requestDescriptor: RequestDescriptor): Promise
 
         RPCResponseObserver.addEventListener(`response:${request.id}`, handler);
     });
+
+    await Transport.send(request);
+
+    console.error('DATA SENT', Date.now());
+
```
d-gubert, line 148:
```suggestion
```

Nope.

PR #736, opened 2024-03-12
Grant networking permission to Deno subprocess

d-gubert:
# What? :boat:

- Grant networking permission to Deno subprocess when applicable
- Fix an issue where Socket comms would end too early
- Small refactor of require function
# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/server/runtime/AppsEngineDenoRuntime.ts
+++ b/src/server/runtime/AppsEngineDenoRuntime.ts
@@ -124,7 +124,19 @@ export class DenoRuntimeSubprocessController extends EventEmitter {
             // process must be able to read in order to include files that use NPM packages
             const parentNodeModulesDir = path.dirname(path.join(appsEngineDir, '..'));
 
-            this.deno = child_process.spawn(denoExePath, ['run', `--allow-read=${appsEngineDir}/,${parentNodeModulesDir}/`, denoWrapperPath, '--subprocess']);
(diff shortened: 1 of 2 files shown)


src/server/runtime/AppsEngineDenoRuntime.ts
```diff
@@ -124,7 +124,19 @@ export class DenoRuntimeSubprocessController extends EventEmitter {
             // process must be able to read in order to include files that use NPM packages
             const parentNodeModulesDir = path.dirname(path.join(appsEngineDir, '..'));
 
-            this.deno = child_process.spawn(denoExePath, ['run', `--allow-read=${appsEngineDir}/,${parentNodeModulesDir}/`, denoWrapperPath, '--subprocess']);
+            let hasNetworkingPermission = false;
+
+            if (!appPackage.info.permissions || appPackage.info.permissions.findIndex((p) => p.name === 'networking.default')) {
```
d-gubert, line 129:
```suggestion
            // If the app doesn't request any permissions, it gets the default set of permissions, which includes "networking"
            // If the app requests specific permissions, we need to check whether it requests "networking" or not
            if (!appPackage.info.permissions || appPackage.info.permissions.findIndex((p) => p.name === 'networking.default')) {
```

tapiarafael (approved):

PR #737, opened 2024-03-12
fix: wrap `room.usernames` with await

tapiarafael:
# What? :boat:

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/deno-runtime/lib/ast/operations.ts
+++ b/deno-runtime/lib/ast/operations.ts
@@ -214,3 +214,29 @@ export const fixLivechatIsOnlineCalls: FullAncestorWalkerCallback<WalkerState> =
 
     state.isModified = true;
 };
+
+export const fixRoomUsernamesCalls: FullAncestorWalkerCallback<WalkerState> = (node, state, ancestors) => {
+    if (node.type !== 'MemberExpression' || node.computed) return;
+
+    if ((node.property as Identifier).name !== 'usernames') return;
+
+    let parentIndex = ancestors.length - 2;
+    let targetNode = ancestors[parentIndex];
+
+    if (targetNode.type !== 'CallExpression') {
+        targetNode = node;
+    } else {
+        parentIndex--;
+    }
+
+    // If we're already wrapped with an await, nothing to do
+    if (ancestors[parentIndex].type === 'AwaitExpression') return;
+
+    // If we're in the middle of a chained member access, we can't wrap with await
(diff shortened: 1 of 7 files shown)


deno-runtime/lib/ast/operations.ts
```diff
@@ -214,3 +214,29 @@ export const fixLivechatIsOnlineCalls: FullAncestorWalkerCallback<WalkerState> =
...
+        targetNode = node;
+    } else {
+        parentIndex--;
+    }
+
+    // If we're already wrapped with an await, nothing to do
+    if (ancestors[parentIndex].type === 'AwaitExpression') return;
+
+    // If we're in the middle of a chained member access, we can't wrap with await
+    if (ancestors[parentIndex].type === 'MemberExpression') return;
```
d-gubert, line 236: I think in this case it is possible to wrap with `await`

For instance, the following code

```js
room.usernames.forEach((name) => console.log(name));
```

produces the following tree (essentially `CallExpression<MemberExpression<MemberExpression<room, usernames>, forEach>>`)

```json
{
  "type": "ExpressionStatement",
  "expression": {
    "type": "CallExpression",
    "callee": {
      "type": "MemberExpression",
      "object": {
        "type": "MemberExpression",
        "object": {
          "type": "Identifier",
          "name": "room"
        },
        "property": {
          "type": "Identifier",
          "name": "usernames"
        },
        "computed": false,
        "optional": false
      },
      "property": {
        "type": "Identifier",
        "name": "forEach"
      },
      "computed": false,
      "optional": false
    },
    "arguments": [
      {
        "type": "ArrowFunctionExpression",
        "id": null,
        "expression": true,
        "generator": false,
        "async": false,
        "params": [
          {
            "type": "Identifier",
            "name": "name"
          }
        ],
        "body": {
          "type": "CallExpression",
          "callee": {
            "type": "MemberExpression",
            "object": {
              "type": "Identifier",
              "name": "console"
            },
            "property": {
              "type": "Identifier",
              "name": "log"
            },
            "computed": false,
            "optional": false
          },
          "arguments": [
            {
              "type": "Identifier",
              "name": "name"
            }
          ],
          "optional": false
        }
      }
    ],
    "optional": false
  }
}
```

and can be modified into 

```js
(await room.usernames).forEach((name) => console.log(name));
```

So I think this validation can be removed

d-gubert (changes requested):

d-gubert (approved):

PR #754, opened 2024-04-12
feat: Apps Engine Bridge Method for Unread Room Messages

Dnouv:
# What? :boat:

- Added bridge for reading room unread messages for a user
- [TBA] Added bridge for counting the unread messages for a user in a room
# Why? :thinking:

# Links :earth_americas:

- [Rocket.Chat Bridge PR](https://github.com/RocketChat/Rocket.Chat/pull/32194)

# PS :eyes:

### DIFF
--- a/src/definition/accessors/IMessageRead.ts
+++ b/src/definition/accessors/IMessageRead.ts
@@ -12,4 +12,25 @@ export interface IMessageRead {
     getSenderUser(messageId: string): Promise<IUser | undefined>;
 
     getRoom(messageId: string): Promise<IRoom | undefined>;
+
+    /**
+     * Retrieves an array of unread messages for a specific user in a specific room.
+     *
+     * @param rid The unique identifier of the room from which to retrieve unread messages.
+     * @param uid The unique identifier of the user for whom to retrieve unread messages.
+     * @param options Optional parameters for retrieving messages:
+     *                - limit: The maximum number of messages to retrieve. If more than 100 is passed, it defaults to 100.
+     *                - skip: The number of messages to skip (for pagination).
+     *                - sort: An object defining the sorting order of the messages. Each key is a field to sort by, and the value is either 1 for ascending order or -1 for descending order.
+     * @returns A Promise that resolves to an array of IMessage objects representing the unread messages for the specified user in the specified room.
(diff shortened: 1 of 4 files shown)


src/definition/accessors/IRoomRead.ts
```diff
@@ -90,4 +90,24 @@ export interface IRoomRead {
...
+     * @returns A Promise that resolves to an array of IMessage objects representing the unread messages for the specified user in the specified room.
+     */
+    getUnreadByRoomAndUser(roomId: string, uid: string, options?: Partial<GetMessagesOptions>): Promise<IMessageRaw[]>;
+
+    /**
+     * Gets the user's unread messages count in a room.
+     * @param uid user's id
+     * @param roomId room's id
+     */
+    getUserUnreadMessageCountByRoom(uid: string, roomId: string): Promise<number | undefined>;
```
d-gubert: In what cases will the return of the count be `undefined`?

Also, I think you can drop the `ByRoom` from both the methods, since they are already in the `IRoomRead` and a room filter can be implied

d-gubert: Another thing, try to be consistent in your parameters: in one of the methods you have `(uid, roomId)` and in the other you have `(roomId, uid)` - it's better to keep a pattern. My suggestion is to keep the `roomId` as the first parameter

d-gubert (changes requested):

src/definition/accessors/IRoomRead.ts
Dnouv: > In what cases will the return of the count be undefined?

Could be in case of any database query or any other unexpected errors.

d-gubert: In that case the method should throw, not return undefined

d-gubert: As a rule of thumb, methods like counts or getMany should only return one type, number and array (respectively); if there is no data, return `0` or `[]`; if there's an error, `throw`

Dnouv: Got it, thank you!

d-gubert (approved):

PR #760, opened 2024-04-29
fix: `OmnichannelSource` type

ggazzo:
# What? :boat:

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:


d-gubert:
I think in this instance we can disregard the potential of breaking changes as it is unlikely anybody would have been able to make a functioning app that was misusing the types here, so we're good to go

PR #778, opened 2024-07-03
feat: `onUpdate` hook for apps

KevLehman:
# What? :boat:

Added hook to be called when an app is updated

# Why? :thinking:

Because why not 🤷🏽‍♂️ 
Serious: allows apps to do stuff only when an update happens, like sending messages or create cron jobs.
# Links :earth_americas:

Related PR to get `user` on context: https://github.com/RocketChat/Rocket.Chat/pull/32719
# PS :eyes:


src/server/AppManager.ts
```diff
@@ -718,6 +718,8 @@ export class AppManager {
                 .catch(() => {});
         }
 
+        await this.updateApp(app)
```
KevLehman, line 721:
```suggestion
        await this.updateApp(app);
```

d-gubert:
Thanks for this! 

I just think we're missing here the signature for the method in the `App` class, like the [`onInstall` event here](https://github.com/RocketChat/Rocket.Chat.Apps-engine/blob/3a59d4018a94026bd269bc569fa31f28a92a5a63/src/definition/App.ts#L182) for example. Regarding contextual information about the event, I'd say it would be cool to know the user who triggered the update and maybe what the previous version was? I'm open to ideas :)

src/definition/accessors/IAppUpdateContext.ts
```diff
@@ -0,0 +1,6 @@
+import type { IUser } from '../users';
+
+export interface IAppUpdateContext {
+    user: IUser | null;
```
d-gubert, line 4: This looks like an optional property, is there a reason why you didn't use `user?: IUser`?

KevLehman: Hmmm, good question. Cannot remember 🤔 let me double check

d-gubert:
```suggestion
    user?: IUser;
```

d-gubert:
Example usage:

```
    public async onUpdate(context: IAppUpdateContext, read: IRead, http: IHttp, persistence: IPersistence, modify: IModify): Promise<void> {
        console.log("Hello from update", context);

        const message = modify.getCreator().startMessage();

        const room = await read.getRoomReader().getById("GENERAL");

        if (!room) {
            throw new Error('oh no');
        }

        message.setText("Hello, thank you for updating!").setRoom(room);

        await modify.getCreator().finish(message);
    }

File: src/server/AppManager.ts [hunk 3]
```diff
@@ -650,7 +650,11 @@ export class AppManager {
         this.apps.delete(app.getID());
     }
 
-    public async update(appPackage: Buffer, permissionsGranted: Array<IPermission>, updateOptions = { loadApp: true }): Promise<AppFabricationFulfillment> {
+    public async update(
+        appPackage: Buffer,
+        permissionsGranted: Array<IPermission>,
+        updateOptions: { loadApp?: boolean; user: IUser | null } = { loadApp: true, user: null },
```
d-gubert, line 656:
```suggestion
        updateOptions: { loadApp?: boolean; user?: IUser } = { loadApp: true },
```

d-gubert (approved):

PR #785, opened 2024-07-24
fix: optional methods in video conf providers not being handled correctly

d-gubert:
# What? :boat:

Similarly to how the `checkPre` methods work on the listener manager, we have optional methods in video conf providers that should not block execution if they are not implemented

# Why? :thinking:

# Links :earth_americas:

# PS :eyes:

### DIFF
--- a/src/server/managers/AppVideoConfProvider.ts
+++ b/src/server/managers/AppVideoConfProvider.ts
@@ -89,8 +89,13 @@ export class AppVideoConfProvider {
 
             return result as string | boolean | Array<IBlock> | undefined;
         } catch (e) {
-            if (method === AppMethod._VIDEOCONF_IS_CONFIGURED && e?.code === JSONRPC_METHOD_NOT_FOUND) {
-                return true;
+            if (e?.code === JSONRPC_METHOD_NOT_FOUND) {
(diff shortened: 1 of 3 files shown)


src/server/managers/AppVideoConfProvider.ts
```diff
@@ -86,8 +87,12 @@ export class AppVideoConfProvider {
                 params: runContextArgs,
             });
 
-            return result as string;
+            return result as string | boolean | Array<IBlock> | undefined;
         } catch (e) {
+            if (method === AppMethod._VIDEOCONF_IS_CONFIGURED && e?.code === JSONRPC_METHOD_NOT_FOUND) {
+                return true;
+            }
+
```
pierre-lehnen-rc, line 95:
```suggestion
            if (e?.code === JSONRPC_METHOD_NOT_FOUND) {
                if (method === AppMethod._VIDEOCONF_IS_CONFIGURED) {
                    return true;
                }
                if (![AppMethod._VIDEOCONF_GENERATE_URL, AppMethod._VIDEOCONF_CUSTOMIZE_URL].includes(method)) {
                    return undefined;
                }
            }

```

Just to avoid logging "method not found" errors for methods that are optional.

pierre-lehnen-rc (approved):

